PR #2, opened 2019-01-09
Minor Updates

FarnazH:
Fix multiple things as described in the commit messages.


FarnazH:
@matt-chan After merge, I think we should tag this to make a new release independent of `gbaiss` and `iodata` under the name `theochem/grid`. Can you please do that?

matt-chan:
> a new release independent of gbaiss and iodata under the name theochem/grid

@FarnazH Do you mean in conda?

FarnazH:
> > a new release independent of gbaiss and iodata under the name theochem/grid
> 
> @FarnazH Do you mean in conda?

@matt-chan yes, a conda release.

PaulWAyers:
Doing nuclear-nuclear repulsion energy in C is sort of absurd. If that is *ever* within a factor of 1000 of the time-limiting step in your calculation it is only because your calculation crashed before it could compute anything else. This is one of the type of obvious thing that should be taken care of when refactoring...it doesn't even require testing to make sure that it is a computational bottleneck.

PR #4, opened 2019-01-11
Remove gbasis & iodata dependency

FarnazH:
The `gbasis` & `iodata` were used in two tests, so the data computed through these two modules were stored in a `.npz` file, and dependency was removed.


matt-chan:
Looks good. Thanks for separating it out @FarnazH!

PR #5, opened 2019-01-21
Rename old_grids to grid

FarnazH:


matt-chan:
Hmm? I thought we merged this in already?

FarnazH:
The only `pylint` error left is `no-name-in-module` which I believe can be ignored. So, this should be good to go.

FarnazH:
> Hmm? I thought we merged this in already?

No. I closed that PR because it was failing on Travis. It is properly done now.

FarnazH:
No, those changes were already merged in an earlier PR. So, we are good to go I think.

matt-chan:
Ah okay. Then I'll merge it.

PR #21, opened 2019-03-01
Remove cell module and tests

FarnazH:
We are not going to support periodic boundary conditions at this
stage, so this cpp implementation can be removed.


tovrstra:
Can you remove corresponding settings in `setup.py`?

FarnazH:
I tried to remove `cell` from `setup.py` and other modules that use it, like `uniform`, `evaluate` and `cext`... It took some time, and not all the tests were passing at the end. I think it's better to have `cell` be removed at a later stage, so I will close this PR.

Issue #8, opened 2019-01-22
1D Grid, e.g. for radial integrals


FarnazH:
`OneDGrid` Class, just a namedtuple with `points` and `weights` attributes. Two categories of functions should be present:

1) **Generation.** Functions to generate an instance of `OneDGrid` with new points and weights from scratch. These functions should have one mandatory argument, i.e. the first argument: the number of grid points. They could take additional arguments to control other aspects of the integration grids. All these functions should have names starting with `generate_`.

2) **Transformation.** Functions that take an existing `OneDGrid` instance as input (first argument) and optionally some parameters to control the transformation. The return value is a new instance of `OneDGrid`. All these functions should have names starting with `transform_`.

Paul's notes and fortran code contain many examples of both types of functions.

No need to make very long function names like `generate_onedgrid_*` and  `transform_onedgrid_*`, because they already are collected in a module that provides a `onedgrid` namespace.

tovrstra:
See theochem/qcgrids#3

PaulWAyers:
Should we have 4*pi*r^{2} or just r^{2}? Because:
* I would expect the 4*pi to show up in the angular integral.
* It might be useful to generalize to r^{n} n >= 0

tovrstra:
I've edited this post and the title to better reflect the discussion we had one how to implement these types of grids. Feel free to comment.

PaulWAyers:
I like having generic 1Dgrid and 1Dtransformation.  That's what it is.

In my notes I suggest supporting several different interpolation strategies, from "fast and dumb" to "clever" with the transformation mapped back in. That means one might need to be able to compute the derivatives of the transformation (or its inverse) at a point.  So the transformation may be something different:  it's something that just reads in points and prints out transformed points and/or derivatives of the transformation. Then if you call that for the whole vector of points you naturally get the new vector of points and the factors you need to multiply the weights by.  (And one can have a porcelain layer that does that multiplication for you.)

PR #22, opened 2019-03-05
Translated rtransform module

tczorro:
1. Tranlate rtransform.cpp to rtransform.py
2. Update test_rtransform for the new pure python version

### DIFF
--- a/grid/grid/rtransform.py
+++ b/grid/grid/rtransform.py
@@ -20,7 +20,6 @@
 # --
 
 # Transformation from uniform 1D to non-uniform 1D grids.
-from abc import abstractmethod
 from numbers import Number
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


tovrstra:
It may not be worth keeping the old RTransform alive, because most of its features will no longer be useful. I still need to take a look at Paul's notes on the integration grids, but as far as I remember, we could drastically simplify the 1D transformations with much less API than we used to have.

tovrstra:
@tczorro: Can you take a look at #8? This is essentially the only thing we need. I guess the features described there could be useful for any type of one-dimensional grid, so we should probably call this `OneDGrid` rather than `RadialGrid`.

FarnazH:
Considering our 3 phase plan for `grid` module makeover, I think @tczorro 's PR is aligned with phase 1, and changes that @tovrstra mentioned would be taken care of in near future. The linter's errors are also excusable at this stage, so I will merge this PR.

PR #31, opened 2019-03-21
Add basic oned grid class with basic functions

tczorro:
Add basic Grid class as OneDGrid with points and weights.
Add a basic module to generate different kind of 1dgrid according to the notes.

### DIFF
--- a/src/grid/grid.py
+++ b/src/grid/grid.py
@@ -28,39 +28,21 @@ class Grid:
             )
         self._points = points
         self._weights = weights
-        self._size = self._points.shape[0]
+        self._size = self._weights.size
 
     @property
     def points(self):
-        """Return the coordinates of grid.
-
-        Returns
-        -------
-        np.ndarray(N,)
-            Coordinates of the grid points
-        """
+        """np.ndarray(N,): the coordinates of each grid point."""
         return self._points
 
     @property
     def weights(self):
-        """Return the weights of each grid.
-
-        Returns
-        -------
-        np.ndarray(N,)
-            Description
-        """
+        """np.ndarray(N,): the weights of each grid point."""
         return self._weights
 
     @property
     def size(self):
-        """Return the total number of points on the grid.
-
-        Returns
-        -------
-        int
-            Total number of grid points
-        """
+        """int: the total number of points on the grid."""
         return self._size
 
     def __getitem__(self, index):
--- a/src/grid/grid.py
+++ b/src/grid/grid.py
@@ -1,0 +1,128 @@
+"""Comstruct basic grid data structure."""
+import numpy as np
+
+
+class Grid:
+    """Basic Grid class for grid information storage."""
+
+    def __init__(self, points, weights):
+        """Construct Grid instance.
+
+        Parameters
+        ----------
+        points : np.ndarray(N,)
+            An array with coordinates as each entry.
+
+        weights : np.ndarray(N,)
+            An array of weights associated with each point on the grid.
+
+        Raises
+        ------
+        ValueError
+            Shape of points and weights does not match.
+        """
+        if len(points) != len(weights):
+            raise ValueError(
+                "Shape of points and weight does not match. \n"
+                "shape of points: {len(poitns)}, shape of weights: {len(weights)}."
+            )
+        self._points = points
+        self._weights = weights
+        self._size = self._points.shape[0]
+
+    @property
+    def points(self):
+        """Return the coordinates of grid.
+
+        Returns
+        -------
+        np.ndarray(N,)
+            Coordinates of the grid points
+        """
+        return self._points
+
+    @property
+    def weights(self):
+        """Return the weights of each grid.
+
+        Returns
+        -------
+        np.ndarray(N,)
+            Description
+        """
+        return self._weights
+
+    @property
+    def size(self):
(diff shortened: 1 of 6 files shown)


tczorro:
Hi @FarnazH , I checked the difference in assert_allclose and allclose. One of the differences is that assert_allclose set atol to 0, which means by default setting, which means for very small target value, it will constantly fail.
Also, I am not totally comfortable mixing numpy with testing modules. So for now, I will stick build-in assert command.

PaulWAyers:
The atol = 0 means that by default *only* the relative error is checked. That's appropriate for checks of numerical precision/round-off since it means it checks that 7 digits are correct.  If you instead use atol > 1e-7 and rtol = 0 it forces the values to be actually similar, and doesn't accept, say, 123456789.012345 and 123456789.012344 as equal (which would be accepted with default values). So in most cases the default parameters are equal (though rtol = 1e-7 may need to be tightened sometimes).

tczorro:
I agree with the point. But it brings problem when the target value is small. For example, any number which is not exact 0 will raise error when you use assert_allclose(1e-30, 0), while allclose is fine which check both arol and rtol. We can always pass our own args to fit specific situation.
I don’t mean no.testing is bad. I just prefer not to use numpy for test purpose but for math purpose.

PaulWAyers:
OK, I understand. For the case you want you'd want assert_allclose(1e-30,0,rtol=1e-7,atol=1e-15) (or something like that). Of course sometimes you would want something to fail if you had assert_allclose(1e-30,0). (It would mostly happnen when you are choosing an ill-chosen set of units, though.)

tczorro:
I just prefer allclose(a, b) since it offers more flexibility (personally). np.allclose returns a boolean value where I can choose to test either case and raise error if I want.
I just came up with another example (feel free to point me out if I am wrong)
>> a = [1, 2, 3, 4, 5]
>> assert np.allclose(a, [1, 2, 3, 4, 5])  # return True
>> b = random_permute(a)
>> assert not np.allclose(b, [1, 2, 3, 4, 5])  # return (not False) Trne

tovrstra:
@tczorro I see your point and I also prefer using the built-in `assert` statement when possible because it plays nicely with syntax highlighting, making unit tests more readable. There is a downside to it too. For example, take the following unit test module:

```python
# test.py

import numpy as np

def test_foo1():
    assert np.allclose(1e-30, 0, atol=0)

def test_foo2():
    np.testing.assert_allclose(1e-30, 0)
```

Both tests will fail. When running `pytest ./test.py`, one gets the following:

```
============================= test session starts ==============================
platform linux -- Python 3.7.2, pytest-4.3.0, py-1.8.0, pluggy-0.9.0
rootdir: /home/toon, inifile:
collected 2 items                                                              

test.py FF                                                               [100%]

=================================== FAILURES ===================================
__________________________________ test_foo1 ___________________________________

    def test_foo1():
>       assert np.allclose(1e-30, 0, atol=0)
E       assert False
E        +  where False = <function allclose at 0x7f17cadddea0>(1e-30, 0, atol=0)
E        +    where <function allclose at 0x7f17cadddea0> = np.allclose

test.py:5: AssertionError
__________________________________ test_foo2 ___________________________________

    def test_foo2():
>       np.testing.assert_allclose(1e-30, 0)
E       AssertionError: 
E       Not equal to tolerance rtol=1e-07, atol=0
E       
E       (mismatch 100.0%)
E        x: array(1.e-30)
E        y: array(0)

test.py:9: AssertionError
=========================== 2 failed in 0.10 seconds ===========================
```

The second error message is more informative, which would get my preference. I try to stick to the following, but I'm still figuring things out, so comments are very welcome:

- `assert_allclose` for testing if (complex) floats and numpy arrays of (complex) floats are similar enough. I'm not so worried about the defaults of `atol` and `rtol`, because they can be overridden when needed.
- `assert_equal` for testing if numpy arrays of non-float types (integers etc.) are the same.
- The built-in `assert` statement for anything else.

For most non-array types, even lists, pytest can generate very nice informative errors, e.g. 

```
    def test_list1():
>       assert [1, 2, 3] == [1, 3, 3]
E       assert [1, 2, 3] == [1, 3, 3]
E         At index 1 diff: 2 != 3
E         Use -v to get the full diff
```

With the `-v` option:

```
>       assert [1, 2, 3] == [1, 3, 3]
E       assert [1, 2, 3] == [1, 3, 3]
E         At index 1 diff: 2 != 3
E         Full diff:
E         - [1, 2, 3]
E         ?     ^
E         + [1, 3, 3]
E         ?     ^
```

This is very clear. Using `assert_equal` for this example would also be ok but both the assert code and the test output are not as nice:

```
    def test_list2():
>       np.testing.assert_equal([1, 2, 3], [1, 3, 3])
E       AssertionError: 
E       Items are not equal:
E       item=1
E       
E        ACTUAL: 2
E        DESIRED: 3
```

The `-v` option has no effect on `assert_equal`.

tczorro:
@tovrstra Good point. The default assert is quite hard to see what went wrong when error happened. This can be very useful when debugging or fixing broken code.
I think I will try it out for my tests when allclose and equal is needed.

tovrstra:
Sounds good. These things will become easier as we get some more mileage.

PR #34, opened 2019-04-04
Add angular grid

tczorro:
Added lebedev angular grid


FarnazH:
All looks good to me. I just have a small question about `lebedev_laikov_lmaxs` dictionary which is removed. Don't we need it?

tczorro:
I merged the dictionary into the method `_select_grid_type`. So It would use a for-loop to search the proper magic number L or number points N for given arg either number of points or the degree of freedom.

PR #53, opened 2019-05-30
[Inactive] Add 6d integral for integral.py

tczorro:
1. Add two grid integral with customized dis, value func.
2. Add two simple tests for integral. More are needed


FarnazH:
@tczorro is this ready to be reviewed?

tczorro:
> @tczorro is this ready to be reviewed?

I'm not sure do we still need it. Let's keep it open and see. I am pretty sure it is obsolete in many aspects and may need to rewrite.

PR #56, opened 2019-06-05
Modifications of GaussLaguerre and GaussChebyshev functions

thomaspigeon:
I wrote some modifications in onedgrid and test_onedgrid so that these functions return a grid which is suited to approximate the integral of f(x) and not f(x)*w(x) where w(x) is the weight function associated to the quadrature (x^alpha * e^(-x) for Gauss Laguerre and 1/sqrt(1 - x^2) for Gauss Chebyshev).

### DIFF
--- a/src/grid/onedgrid.py
+++ b/src/grid/onedgrid.py
@@ -83,11 +83,16 @@ def GaussLegendre(npoints):
 
 
 def GaussChebyshev(npoints):
-    r"""Generate Gauss-Chebyshev grid.
+    r"""Generate a grid based on Gauss-Chebyshev grid.
 
+    Gauss Chebyshev grid is defined as:
     .. math::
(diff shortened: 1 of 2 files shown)


src/grid/onedgrid.py
```diff
@@ -33,16 +33,24 @@ def generate_onedgrid(npoints, *args):
...
 
+    Generalizeed Gauss Laguerre grid is defined as:
     .. math::
-        \int_{0}^{\infty} e^{-x} f(x)dx \approx \sum_{i=1}^n w_i f(x_i)
+        $$\int_{0}^{\infty} x^{\alpha}e^{-x} f(x)dx \approx
+        \sum_{i=1}^n w_i f(x_i)
+
+    This integration grid is defined as :
+    .. math::
+        $$\int_{0}^{\infty} f(x)dx \approx
```
tczorro: Sphinx use `.. math::` as math directive. So it would be better to remove the `$$` sign.

PR #63, opened 2019-07-05
Refactoring 

FarnazH:
This PR is meant to share some of the changes @tovrstra and I was thinking about. I will be pushing more commits in the coming days. Please let us know what you think.


PaulWAyers:
My only reluctance about this renaming is that in practice the 1D grids and transformed 1D grids in `grid` have validity and utility for general-purpose 1D integration, not just for radial integrals.

FarnazH:
@PaulWAyers you are right, probably it was not the best choice and it would have been better to for all of us to talk about names before making any changes. I will try to avoid renaming classes/functions/attributes until we can have a joint meeting.

PR #64, opened 2019-07-06
Refactor

FarnazH:
Please see the commits below, and feel free to let me know what you think.

### DIFF
--- a/src/grid/rtransform.py
+++ b/src/grid/rtransform.py
@@ -19,6 +19,8 @@
 #
 # --
 """Transformation from uniform 1D to non-uniform 1D grids."""
+
+
 import warnings
 from abc import ABC, abstractmethod
 from numbers import Number
@@ -32,23 +34,23 @@ class BaseTransform(ABC):
     """Abstract class for transformation."""
 
     @abstractmethod
-    def transform(self, array):
+    def transform(self, r):
         """Abstract method for transformation."""
 
     @abstractmethod
-    def inverse(self, r_array):
+    def inverse(self, r):
         """Abstract method for inverse transformation."""
 
     @abstractmethod
-    def deriv(self, array):
+    def deriv(self, r):
         """Abstract method for 1st derivative of transformation."""
 
     @abstractmethod
-    def deriv2(self, array):
+    def deriv2(self, r):
         """Abstract method for 2nd derivative of transformation."""
 
     @abstractmethod
-    def deriv3(self, array):
+    def deriv3(self, r):
         """Abstract method for 3nd derivative of transformation."""
 
     def generate_radial(self, oned_grid):
@@ -165,12 +167,12 @@ class BeckeTF(BaseTransform):
             mid_value = (array[size // 2 - 1] + array[size // 2]) / 2
         return (radius - r0) * (1 - mid_value) / (1 + mid_value)
 
-    def transform(self, array, trim_inf=True):
+    def transform(self, r, trim_inf=True):
         """Transform given array[-1, 1] to radial array[r0, inf).
 
         Parameters
         ----------
-        array : np.ndarray(N,)
+        r : np.ndarray(N,)
             One dimension numpy array located between [-1, 1]
         trim_inf : bool, optional, default to True
             Flag to trim infinite value in transformed array. If true, will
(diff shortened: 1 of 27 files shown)


FarnazH:
@tczorro I went back to each commit and fixed all `black` and `flake8` errors, so the history is clean now for you to review this PR. Thanks.

src/grid/rtransform.py
```diff
@@ -32,23 +34,23 @@ class BaseTransform(ABC):
     """Abstract class for transformation."""
 
     @abstractmethod
-    def transform(self, array):
+    def transform(self, r):
         """Abstract method for transformation."""
 
     @abstractmethod
-    def inverse(self, r_array):
+    def inverse(self, r):
```
tczorro: inverse r and transform r are different, so I would prefer them to be different arg name

FarnazH: OK, is it good to use `r` for transform/deriv/deriv2/deriv3 methods and `r_array` for inverse?

tczorro: it doesn't matter what name to use. We can use x for [-1, 1] and r for [0, inf]. But mixing these two i think is confusing for people who read the code.

FarnazH: I agree it is better not to mix things. I liked `x` better to be honest, but as we are using these 1D grids as a radial grid mostly, I thought `r` would be better.

FarnazH: @tovrstra do you have any suggestions?

tczorro: oned(x) need to be transform to generate radial(r). So i think it's ok.
TF:
transform(x) -> r
inverse(r) -> x
INV:
transform(r) -> x
inverse(x) -> r

src/grid/tests/test_onedgrid.py
```diff
@@ -1,61 +1,71 @@
...
-        """Test horton linear grids."""
-        grid = HortonLinear(10)
-        assert_allclose(grid.points, np.arange(10))
-        assert_allclose(grid.weights, np.ones(10))
-
-    def test_errors_raise(self):
-        """Test errors raise."""
-        with self.assertRaises(ValueError):
-            GaussLaguerre(10, -1)
+def test_gauss_laguerre():
```
tczorro: I'd like it to be a unittest class more. It's a buildint test class with some nice feature like `mock` etc, which we probably can use in the future. It also keeps the tests file style consistent.

src/grid/rtransform.py
FarnazH: @tczorro see 63bc25c, let me know if anything is left out. Do you know why `flake8` is failing?

src/grid/tests/test_onedgrid.py
FarnazH: Let's talk about this later. I think it is important to be consistent across modules (iodata, gbasis, and chemtools all use the test functions), and unless we are using the extra features of the test classes that you were referring to, I am not sure we should bother with them (as they look/seem/are more complicated and the extra indentation is a waste of line characters). Also, some of those features might be achievable with pytest. That's just my opinion, but I am not an expert. We should wait until we talk with @tovrstra and also see what @kimt33 has to say.

tovrstra: I believe the `unittest.TestCase` class has little added value when using pytest. When running tests with `python -m unittest <modulename>`, I believe the classes are mandatory and one should also be using its methods like `self.assertTrue(...)`, etc. The main goal of frameworks like pytest (and previously nosetests) is to make unit testing easier, e.g. without `TestCase` classes and by encouraging simple usage of the `assert` statement instead of the large number of `TestCase.assert*` methods.

Also the pytest fixtures are easier to use than the `setUp` and `tearDown` methods of the `TestCase` class. See for example the following test in iodata, which needs a temporary directory: https://github.com/theochem/iodata/blob/master/iodata/test/test_cube.py#L99
Just by adding `tmpdir` argument, the setup and teardown of the temporary directory are taken care of. With the old unittest module, this would require one to write `setUp` and `tearDown` methods. I've also used pytest fixtuyres in tinydft to parameterize tests, which worked very well. 

The `unittest.mock` module can be used also without the `TestCase` class, see e.g. examples here: https://docs.python.org/dev/library/unittest.mock.html#quick-guide.  The pytest-mock plugin is designed to make mock easier to use with pytest, see https://pypi.org/project/pytest-mock/

If there are other features of the `TestCase` class, please let me know. As far as I can tell, it should be easier to write unit tests without these classes, without losing features of the old `unittest` module.

tczorro: @tovrstra Thank you very much for the explanation. I do agree `pytest` work like a charm for testing and I like it from the very beginning.
`Unittest` are `pytest` are not exclusive to each other. We can combine both to get all the edgy features if we need. 
So here's my defence for not to change this test file.

1. `unittest` is a buildin module. If the feathers needed are offered in `unittest`, there's little motivation for me resorting to a 3rd party module. (though I put `pytest` in setup.py, it's not really a dependence).

2. For `Grid`, tests are frame free. One can use `pytest`, `nosetests`, or even 'unittest' to run all the tests without problem. User/developer are free to use their preferred test framework.

3. To support `unittest` style test, it only need one extra line. And to me, it doesn't look terribly ugly or complicated.

tczorro:
PSA: The flake8 error is due to flake8-docstrings incompatibility with new version of pydocstyle. It should be fixed soon.

FarnazH:
Based on our conversation, I will break down this PR to smaller ones for review and merge.

PR #68, opened 2019-07-08
Cleanups

FarnazH:
This includes minor cleanups and updates across various modules. Most of them were reviewed in the PR #64. Please let me know if anything needs to be fixed.


FarnazH:
@tczorro The failure is due to flake8-docstrings and pydocstyle conflict, so it's good to be reviewed I think.

PR #71, opened 2019-07-10
Refactor Molgrid & BeckeWeights

FarnazH:
This is not perfectly polished probably, but I wanted to share it asap to get your opinion before spending more time on it.

### DIFF
--- a/src/grid/becke.py
+++ b/src/grid/becke.py
@@ -22,21 +22,17 @@
 
 import warnings
 
-from grid.utils import get_cov_radii
-
 import numpy as np
 
 
 class BeckeWeights:
     """Becke weights functions holder class."""
 
-    def __init__(self, atom_coors, atom_nums, radii=None, order=3):
+    def __init__(self, radii, order=3):
         r"""Initialize class.
 
         Parameters
         ----------
-        atom_coors : np.ndarray(M, 3)
-            Cartesian coordinates of :math:`M` atoms in molecule.
         atom_nums : np.ndarray(M,)
             Atomic number of :math:`M` atoms in molecule.
         radii : np.ndarray(M,), optional
@@ -45,24 +41,9 @@ class BeckeWeights:
             Order of iteration for switching function.
 
         """
-        if atom_coors.ndim != 2 or atom_coors.shape[1] != 3:
-            raise ValueError(
-                f"atom_coors needs to be in shape (M, 3), got {atom_coors.shape}"
-            )
-        if len(atom_coors) != len(atom_nums):
-            raise ValueError(
-                f"atom_coors & atom_nums represent different number of atoms."
-            )
-        if radii is not None and len(atom_coors) != len(radii):
-            raise ValueError(
-                f"The number of atoms in atom_coors and radii do not match."
-            )
         if not isinstance(order, int):
             raise ValueError(f"order should be an integer, got {type(order)}")
 
-        if radii is None:
-            radii = get_cov_radii(atom_nums)
-        self._atom_coors = atom_coors
         self._radii = radii
         self._order = order
 
@@ -117,13 +98,15 @@ class BeckeWeights:
             x = 1.5 * x - 0.5 * x ** 3
         return x
 
-    def generate_weights(self, points, *, select=None, pt_ind=None):
+    def generate_weights(self, points, atom_coords, *, select=None, pt_ind=None):
         r"""Calculate Becke integration weights of points for select atom.
 
         Parameters
         ----------
         points : np.ndarray(N, 3)
             Cartesian coordinates of :math:`N` grid points.
+        atom_coords : np.ndarray(M, 3)
+            Cartesian coordinates of :math:`M` atoms in molecule.
         select : list or integer, optional
             Index of atom index to calculate Becke weights
         pt_ind : list, optional
@@ -139,7 +122,7 @@ class BeckeWeights:
         # |r_A - r| for each points, nucleus pair
         # check ``select``
         if select is None:
-            select = np.arange(len(self._atom_coors))
+            select = np.arange(len(atom_coords))
         elif isinstance(select, (np.integer, int)):
             select = [select]
         # check ``pt_ind`` points index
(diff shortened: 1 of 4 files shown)


src/grid/becke.py
```diff
@@ -19,14 +19,53 @@
...
 
+from grid.utils import get_cov_radii
+
 import numpy as np
 
 
 class BeckeWeights:
     """Becke weights functions holder class."""
 
+    def __init__(self, atom_coors, atom_nums, radii=None, order=3):
```
tovrstra: I would suggest to have only radii and order as constructor arguments, which define the method. Another method can be used to apply it on any given molecular geometry and grid.

src/grid/molgrid.py
```diff
@@ -18,36 +18,32 @@
...
             list of atomic grid
-        radii : np.ndarray(N,)
-            Radii for each atom in the molecular grid
-        aim_weights : str or np.ndarray(K,), default to "becke"
-            Atoms in molecule weights. If str, certain function will be called
-            to compute aim_weights, if np.ndarray, it will be treated as the
-            aim_weights
+        aim_weights : BeckeWeights or np.ndarray(K,)
+            Atoms in molecule weights.
+
```
tovrstra: It may convenient to still support the string API, for backward compatibility and ease of use.

src/grid/becke.py
```diff
@@ -78,37 +117,29 @@ def _switch_func(x, order=3):
             x = 1.5 * x - 0.5 * x ** 3
         return x
 
-    @staticmethod
-    def generate_becke_weights(
-        points, radii, atom_coors, *, select=None, pt_ind=None, order=3
-    ):
-        """Calculate Becke weights of points for select atom.
+    def generate_weights(self, points, *, select=None, pt_ind=None):
```
tovrstra: Could become `__call__`

src/grid/becke.py
tczorro: ``coors`` is also an attribute in ``molgrid``, I think it can be reused to make this initializer simpler.

FarnazH: OK, I agree. I will move `atom_coors` back to `generate_weights` method to make the class general. Do you think the `radii` should become a dictionary? I mean, if `radii` is just a list/array and the order of atoms in the molecule are different, we might use the wrong radius for an atom... Let me know what you think, please.

src/grid/molgrid.py
FarnazH: I was thinking that supporting `str` as an `aim_weights` is more like porcelain, so we can probably support it in a higher-level code.

src/grid/becke.py
FarnazH: Sounds good to me. Then the `HirshfeldWeights` class would have a `__call__` method as well... I was also thinking to put the chunking hack for computing Becke weight in the `__call__` function too so that it would become simpler...

src/grid/becke.py
tczorro: We can put the ``numbers`` back to ``Molgrid`` constructor. This may need some extra structure change.
We may need to add number / radius to atomic grid and extract that information when constructing molgrid. In that case, both radius and coors can be offered directly without manually passing to becke constructor.
In Paul's note, we need radius of the atom to construct proper atomic grid, so it need to be done eventually.

FarnazH: That's a good point and probably the best solution. I will differ it to another PR if that is OK with you, as it touches `AtomicGrid`...

tczorro: I will leave this pr open. Once the ``AtomicGrid`` ready, we can merge both.

FarnazH: The only caveat I see with what you suggested is that we need to pass `radii` (retrieved from `AtomicGrid` instances) when calling `aim_weights`, which wouldn't be needed for HirshfeldWeights` class... So, for the sake of consistent API, I think we might need to pass `radii` to BeckeWeights constructor... @tczorro & @tovrstra what do you think?

tczorro: Extra args can be dump to ``**kwargs`` and poped in its own`` __call__`` function when needed. But this may not looks as elegant as uniform args. I'm not sure will we add new partition schemes in the future. Those schemes may some need extra args.
What I mentioned above are just some extra concerns. For now, I am fine with either way.

PaulWAyers: The atomic radius is used not only for Becke partitioning, but in general in order to specify the "center" of the radial grids (i.e., the value of the radius that is assigned to x=0 in the original -1 <= x <= +1 grid, or analogously, the point corresponding to the "halfway" gridpoint in Gauss-Laguerre 1D-quadrature.

tovrstra: As long as the radii are optional in the constructor of the BeckeWeights class, all is fine. We could support list, array and dictionary, but this can also be changed added later, does not need to be done in this PR. It's up to you. The main point is to fix the mandatory arguments at this stage. 

If the atomic grids would not have a radius parameter, e.g. when using grids from other codes, and when they are not provided as optional argument, one can always fall back to a sensible default.

src/grid/molgrid.py
tovrstra: agreed

src/grid/becke.py
tovrstra: agreed

src/grid/becke.py
```diff
@@ -27,6 +28,25 @@
 class BeckeWeights:
     """Becke weights functions holder class."""
 
+    def __init__(self, radii, order=3):
```
tovrstra: In line with the other discussion, it would be good to make radii optional. If not provided, a series of fallbacks can be used to construct them:
1) Take them from the atomic grids, if they have a radius defined. (There are scenarios when this is not the case.)
2) Use some default radius for each chemical element. This implies that atomic numbers are passed to the `generate_weights` method. This argument would also be needed for the Hirshfeld partitioning.

tovrstra: Having a dictionary for the radii would make the most sense. You want to avoid any molecule-specific information to go into the constructor, such that the same definition of the weights can be used to different molecules.

tovrstra:
I would suggest to make the non-optional part of the API of the methods final before merging.

src/grid/becke.py
```diff
@@ -22,28 +22,36 @@
...
 
         """
         if not isinstance(order, int):
             raise ValueError(f"order should be an integer, got {type(order)}")
-
+        if radii is not None:
+            if not isinstance(radii, dict):
+                raise TypeError(f"radii should be a dictionary, got {type(radii)}")
+            if not np.all([isinstance(k, int) for k in radii.keys()]):
+                raise TypeError(f"radii keys should be integers.")
```
tczorro: If an extra dict for radii is provided, it may update the default radii dict. So the user don't have to offer every atomic radii.

FarnazH: They don't have to provide a radius for every atomic number, just the atomic number they would use later when they call `BeckeWeight` instance is enough. But, if `radii=None`, by default all atomic radii are loaded.

What do you mean by updating the dictionary? Are you talking about new functionality?

tczorro: I assume i have a molecule with 10 different atoms, but I only want to customize one's radius while all the others with default value. In this case, I have to go check the default radii for all the other atoms and put them in the dict.
What I mean is it generate a dict for all atoms, and update the value from arg radii if the user intend to customize.

FarnazH: That's a very good point @tczorro, I added a small patch to include this feature. Thanks.

PR #72, opened 2019-07-11
Make MolGrid weights consistent with Grid class

FarnazH:
1. This makes weight attr contain the total integration weight.
2. MolGrid can now use integrate method of Grid class.
3. AtomicGrid weights are still stored in _atweights attr.


FarnazH:
Good point. Especially if `store=True`, the `atweights` are also stored in `_atomic_grids` attribute. We need to decide what to keep, but I didn't want to make that decision by myself.

tczorro:
For memory management, that's something we need to do in the future comprehensively. Im fine with merging this one. Let's see toon's opinion first.

PR #75, opened 2019-09-18
Doc update

tczorro:
Remove grid in onedgrid and transform files
Add negative inf value check
Update math equations in transform

### DIFF
--- a/src/grid/rtransform.py
+++ b/src/grid/rtransform.py
@@ -53,7 +53,12 @@ class BaseTransform(ABC):
         """Abstract method for 3nd derivative of transformation."""
 
     def generate_grid(self, oned_grid):
-        """Generate a new integral grid by transforming given the OneDGrid.
+        r"""Generate a new integral grid by transforming given the OneDGrid.
+
+        .. math::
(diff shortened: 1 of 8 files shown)


src/grid/rtransform.py
```diff
@@ -52,8 +52,8 @@ def deriv2(self, x):
     def deriv3(self, x):
         """Abstract method for 3nd derivative of transformation."""
 
-    def generate_radial(self, oned_grid):
-        """Generate a radial grid by transforming the OneDGrid.
+    def generate_grid(self, oned_grid):
```
PaulWAyers: Maybe call it "trans1d_grid" or something to make it clear that the grid here is the transformation of the oned_grid you read in? Of course it is (also) a 1d grid....

PaulWAyers: or call it generate_tfgrid or generate_tf1dgrid to harmonize with other nomenclature?

src/grid/rtransform.py
```diff
@@ -53,7 +53,12 @@ def deriv3(self, x):
         """Abstract method for 3nd derivative of transformation."""
```
PaulWAyers: 3rd (not 3nd)

PR #78, opened 2019-09-28
Add a new becke weight function

tczorro:
1. Separate q P_A calculation from becke weights
    evaluation
2. Add a new compute_weight outer loop for compute
    becke weights for each section of points


FarnazH:
@tczorro can you please elaborate what is this PR about? This relates to next week's code review, so it would be good to sort it out before that.

tczorro:
Add a high level function to compute the becke weight of a given atom
Then the molecular becke weight computation will be broken down to compute for each atom

PR #79, opened 2019-10-03
OneDimensionGrid

rayhe88:
Pull request for Onedgrid for the next quadrature / rules, whit their respective test
- GaussChebyshevType2
- GaussChebyshevLobatto
- Trapezoidal
- RectangleRuleSineEndPoints
- RectangleRuleSine
- TanhSinh

### DIFF
--- a/src/grid/onedgrid.py
+++ b/src/grid/onedgrid.py
@@ -24,7 +24,7 @@ from grid.basegrid import OneDGrid
 
 import numpy as np
 
-from scipy.special import roots_genlaguerre
+from scipy.special import roots_chebyu, roots_genlaguerre
 
 
 def GaussLaguerre(npoints, alpha=0):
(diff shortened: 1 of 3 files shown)


src/grid/onedgrid.py
```diff
@@ -114,6 +114,294 @@ def GaussChebyshev(npoints):
...
+
+    Returns
+    -------
+    OneDGrid
+        A 1D grid instance.
+    """
+    if npoints % 2 == 0:
+        raise ValueError("npoints must be odd, given {npoints}")
+
+    jmin = (int)(1 - npoints) / 2
```
tczorro: Everything else looks good to me except that I am not sure what ``(int)`` does here.

rayhe88: Thanks, well I wrote (int) because I was thinking in python 3.6, in python 2.7 it's not necessary, sorry was my mistake, jmin is a counter.

PR #84, opened 2020-01-22
Add indices property to subgrid

tovrstra:
These minor changes were useful/needed when prototyping denspart.

### DIFF
--- a/src/grid/tests/test_molgrid.py
+++ b/src/grid/tests/test_molgrid.py
@@ -348,33 +348,33 @@ class TestMolGrid(TestCase):
         )
         grid = MolGrid([atg1], BeckeWeights(), np.array([1]), store=False)
         fn = np.exp(-2 * np.linalg.norm(grid.points, axis=-1))
-        assert np.allclose(grid.integrate(fn), np.pi)
+        assert_allclose(grid.integrate(fn), np.pi)
         # conventional subgrid
         subgrid = grid.get_subgrid(coords, 12.0)
         subfn = np.exp(-2 * np.linalg.norm(subgrid.points, axis=-1))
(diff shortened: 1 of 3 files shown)


src/grid/tests/test_molgrid.py
```diff
@@ -349,11 +349,19 @@ def test_get_subgrid_1s(self):
...
         fn = np.exp(-2 * np.linalg.norm(grid.points, axis=-1))
         assert np.allclose(grid.integrate(fn), np.pi)
+        # conventional subgrid
         subgrid = grid.get_subgrid(coords, 12.0)
-        fn = np.exp(-2 * np.linalg.norm(subgrid.points, axis=-1))
+        subfn = np.exp(-2 * np.linalg.norm(subgrid.points, axis=-1))
         assert subgrid.size < grid.size
         assert subgrid.size == 10560
-        assert np.allclose(subgrid.integrate(fn), np.pi)
+        assert np.allclose(subgrid.integrate(subfn), np.pi)
```
tczorro: Just leave a note for us. I think we agreed to use ``np.assert_allclose`` to replace redundant assert `np.alllcose`. So one point i will change all the assert command to compatible numpy.test command for better error message

tovrstra: ok, I was indeed not sure anymore. I'll change the new ones now.

tczorro: Lol. It's too trivial anyway. There're some legacy tests with these old style. I think I am going to merge this PR. And maybe later we can reformat the tests into a more consistent style.

tovrstra:
Thanks for the quick action!

PR #86, opened 2020-01-23
Improve docstring of grid base class and improve get_subgrid

tovrstra:
A bunch of unit tests were added to make sure get_subgrid works as expected in various scenarios.

I've also remove self._size and the way it was used in MolGrid. Constructors of subclasses should not use or change attributes defined by the parent class.


tovrstra:
Thanks for the quick merge!

PR #94, opened 2020-03-06
Vectorize some one dimensional Grids

rayhe88:
The next one dimensional grids were modified to avoid for-loops.
- Simpson
- ClenshawCurtis
- FejerFirst
- FejerSecond
- TanhSinh
- RectangleRules

Some of this grids  only one of the two for-loops was avoided


src/grid/onedgrid.py
```diff
@@ -327,17 +327,15 @@ def RectangleRuleSineEndPoints(npoints):
...
 
     for i in range(0, npoints):
-        elements = np.zeros(npoints)
-        elements = np.sin(index_m * np.pi * points[i])
-        elements *= (1 - np.cos(index_m * np.pi)) / (index_m * np.pi)
-
-        weights[i] = (2 / (npoints + 1)) * np.sum(elements)
+        weights[i] = np.sum(
+            np.sin(m * np.pi * points[i]) * (1 - np.cos(m * np.pi)) / (m * np.pi)
+        )
```
tczorro: Here is still a ``for`` loop. Could you please vectorize the way weights are computed so that all weights are computed at once.

src/grid/onedgrid.py
```diff
@@ -384,22 +382,17 @@ def RectangleRuleSine(npoints):
...
-        elements /= index_m
-
-        weights[i] = (4 / (npoints * np.pi)) * np.sum(elements)
-
-        weights[i] += (
-            (2 / (npoints * np.pi ** 2))
-            * np.sin(npoints * np.pi * points[i])
-            * np.sin(npoints * np.pi / 2) ** 2
+        weights[i] += (4 / (npoints * np.pi)) * np.sum(
+            np.sin(m * np.pi * points[i]) * np.sin(m * np.pi / 2) ** 2 / m
```
tczorro: The coronavirus has hit France’s national assembly with one deputy in the hospital, and a member of staff having also been diagnosed.

tczorro: Here is also the place where you could use vectorization skill to get rid of ``for`` loop.

PR #103, opened 2020-08-04
single and double exponential

rayhe88:
Add single and double exponential rules, these are related to tanh-sinh quadrature.

- ExpSinh       (0, infinity)
- LogExpSinh (0, infinity)
- ExpExp        (0, infinity)
- SingleTanh   (-1,1)
- SingleExp     (0, infinity)
- SingleArcSinhExp (0, infinity)


FarnazH:
@rayhe88 thanks for your PR. There are multiple merges (from `theochem/master`, etc. branches) into your`double_exp` branch. That is not desirable, because it makes reviewing your code hard and merging to `theochem/master` wouldn't look nice... If your branch is behind `theochem/master`, you can either rebase or just let it be behind master... Clean you please cleanup your branch before we review it? Let me know if you need any help doing it.

PaulWAyers:
Add @tczorro 's integral test. 

In the long run, we should come up with better default parameters for these grids. I have notes on this and will take a crack at it.

In the docstring, we should be very clear that these methods' performance requires very good choices of parameters. In the web/online documentation, we should mention that these methods are deprecated.

Ali-Tehrani:
This pull-request got replaced with #163 and will now be closed

Issue #101, opened 2020-07-29
Negative Radial Grid Points


FarnazH:
The following example will give this error: `TypeError: Smallest rgrid.points is negative, got -4.4999`
This error is not very helpful, and probably the `rgrid` should have raised an error not `MolGrid`.

```python
import numpy as np

from grid.molgrid import MolGrid
from grid.becke import BeckeWeights
from grid.onedgrid import HortonLinear
from grid.rtransform import BeckeTF


atnums = np.array([8, 1, 1])
atcoords = np.array([[0.0, 0.0, 0.0], [0.0, 0.0, 1.79523983], [1.69257101, 0.0, -0.59840635]])

oned = HortonLinear(100)
rgrid = BeckeTF(1.e-4, 1.5).transform_1d_grid(oned)
grid = MolGrid.horton_molgrid(atcoords, atnums, rgrid, 110, BeckeWeights())
```

FarnazH:
The `print(np.min(rgrid.points), np.max(rgrid.points))` gives `-4.4999 1e+16`.

PR #104, opened 2020-08-05
Add domain & codomain to transform

tczorro:
1. Add domain and codomain property to each transform
2. Change tests accordingly

### DIFF
--- a/src/grid/rtransform.py
+++ b/src/grid/rtransform.py
@@ -53,16 +53,16 @@ class BaseTransform(ABC):
         """Abstract method for 3rd derivative of transformation."""
 
     @property
-    def tf_domain(self):
+    def domain(self):
         """tuple: Transformation domain."""
-        return self._tf_domain
+        return self._domain
 
     @property
-    def tf_codomain(self):
+    def codomain(self):
         """tuple: Transformation codomain."""
-        return self._tf_codomain
+        return self._codomain
 
-    def transform_1d_grid(self, oned_grid):  # TODO: add transform check for domains
+    def transform_1d_grid(self, oned_grid):
         r"""Generate a new integral grid by transforming given the OneDGrid.
 
         .. math::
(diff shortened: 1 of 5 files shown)


src/grid/rtransform.py
```diff
@@ -52,7 +52,17 @@ def deriv2(self, x):
...
-    def transform_1d_grid(self, oned_grid):
+    @property
+    def tf_domain(self):
+        """tuple: Transformation domain."""
+        return self._tf_domain
+
+    @property
+    def tf_codomain(self):
+        """tuple: Transformation codomain."""
+        return self._tf_codomain
```
FarnazH, line 63: Isn't it simpler ans easier to remember to just call them `domain` and `codomain`?

src/grid/rtransform.py
```diff
@@ -77,6 +87,15 @@ def transform_1d_grid(self, oned_grid):
         """
         if not isinstance(oned_grid, OneDGrid):
             raise TypeError(f"Input grid is not OneDGrid, got {type(oned_grid)}")
+        # check domain
+        if (
+            oned_grid.domain[0] < self.tf_domain[0]
+            or oned_grid.domain[1] > self.tf_domain[1]
+        ):
+            raise ValueError(
+                "Input grid domain do not match with tf domain.\n"
```
FarnazH, line 96: To be more clear for the user, we can say:
"Given 1D grid domain does not match the transformation domain"

FarnazH:
Thanks @tczorro. Please see my minor comments above. This PR address issue #101 by raising an informative error when transforming the 1D grid.

src/grid/rtransform.py
tczorro: I also prefer `domain` and `codomain`. Just not sure would it bring ambiguity with the `domain` in grid.

tczorro:
Just to double check, if no further question, i will merge this PR tomorrow (Aug 12th) in case the changes become obsolete.

FarnazH:
Thanks for the update @tczorro. I am have no further comments.

PR #108, opened 2020-08-27
Hirshfeld weights

FarnazH:
This is the first draft of the Hirshfeld weight  implementation.


FarnazH:
@tczorro, black complains about `src/grid/atomgrid.py` file, but I haven't touched that in this PR.

tczorro:
It's fine, just trivial style stuff, I will fix that later.

PR #115, opened 2021-04-14
Update AtomGrid

FarnazH:
The current changes update the docstrings of `AtomGrid` to make them complete and consistent. In addition, a few argument names are changes to add clarity. Before making any further changes, I wanted to clarify a few things regarding the `AtomGrid` constructors which are exemplifiedd below:

```python
from grid.atomgrid import AtomGrid
from grid.onedgrid import GaussChebyshev
from grid.rtransform import BeckeTF

oned = GaussChebyshev(30)
rgrid = BeckeTF(1.e-4, 1.5).transform_1d_grid(oned)

# option 1 (using degrees of Levedev grid)
g = AtomGrid(rgrid, degrees=[11])

# option 2 (using atomic number & preset)
g = AtomGrid.from_predefined(rgrid, 1, 'coarse')

# option 2 (using atomic radius & sectors)
g = AtomGrid.from_pruned(rgrid, 0.5, sectors_r=[0.5, 1., 1.5], sectors_degree=[3, 7, 5, 3])
```

1. Considering the fact that the `AtomGrid.from_predefined` loaded precomputed `sectors_r` & `sectors_size`, shouldn't it just call `AtomGrid.from_pruned` upon loading the sectors information? This is not how `AtomGrid.from_predefined` is currently implemented. So, I am a bit confused.
2. In `HORTON2`, the `coarse` built-in grid specifies the radial grid points as well. See: http://theochem.github.io/horton/2.1.1/tech_ref_grids.html?highlight=insane This is not the case for `AtomGrid`, so how are we going to support `HORTOn2`-type built-in grids?

Any comments are welcomed @tczorro, @tovrstra, and @PaulWAyers.

### DIFF
--- a/src/grid/atomgrid.py
+++ b/src/grid/atomgrid.py
@@ -107,28 +107,20 @@ class AtomGrid(Grid):
         self._size = self._weights.size
 
     @classmethod
-    def from_predefined(
-        cls,
-        atnum,
-        rgrid,
-        grid_type,
-        *_,
-        center=None,
-        rotate=False,
-    ):
+    def from_predefined(cls, rgrid, atnum, grid_type, *_, center=None, rotate=False):
         """High level to construct prefined atomic grid.
 
         Examples
         --------
         # construct an atomic grid for H with fine grid setting
-        >>> atgrid = AtomGrid.from_predefined(1, rgrid, "fine")
+        >>> atgrid = AtomGrid.from_predefined(rgrid,1,"fine")
 
         Parameters
         ----------
-        atnum : int
-            atomic number of atomic grid
         rgrid : RadialGrid
             points where sperical grids will be located
+        atnum : int
+            The atomic number corresponding to the grid.
         grid_type : str
             different accuracy for atomic grid
             include: 'coarse', 'medium', 'fine', 'veryfine', 'ultrafine', 'insane'
--- a/src/grid/atomgrid.py
+++ b/src/grid/atomgrid.py
@@ -209,7 +209,9 @@ class AtomGrid(Grid):
             else np.asarray(center, dtype=float)
         )
         cls._input_type_check(rgrid, center)
-        degrees = cls._generate_degree_from_radius(rgrid, radius, sectors_r, sectors_degree)
+        degrees = cls._generate_degree_from_radius(
+            rgrid, radius, sectors_r, sectors_degree
+        )
         return cls(rgrid, degrees=degrees, center=center, rotate=rotate)
 
     @property
--- a/src/grid/atomgrid.py
+++ b/src/grid/atomgrid.py
@@ -37,31 +37,29 @@ from scipy.spatial.transform import Rotation as R
 class AtomGrid(Grid):
     """Atomic grid construction class."""
 
-    def __init__(
-        self,
-        rgrid,
-        *,
-        degs=None,
-        size=None,
-        center=None,
-        rotate=False,
-    ):
+    def __init__(self, rgrid, *, degrees=None, sizes=None, center=None, rotate=False):
         """Construct atomic grid for given arguments.
 
         Parameters
         ----------
         rgrid : OneDGrid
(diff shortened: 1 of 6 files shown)


src/grid/atomgrid.py
```diff
@@ -92,52 +90,48 @@ def __init__(
...
-    def from_predefined(
-        cls,
-        atnum,
-        rgrid,
-        grid_type,
-        *_,
-        center=None,
-        rotate=False,
-    ):
+    def from_predefined(cls, rgrid, atnum, preset, *_, center=None, rotate=False):
```
tczorro, line 112: May I know the reason to swap the position of `radial grid` and `atomic number`

tczorro:
1. You are right. It should just call `AtomGrid.from_pruned` to construct the instance. I think I wrote that code just to double check the result are consistent between the two methods incase errors or bugs in one or the other.

2. We still support the same grid defined in HT2 as long as the radial is provided by the user. When I design grid, I believe, the new radial grids we implemented gonna perform better than the old `power` or `exp` radial. So it's less meaningful to offer an inferior default choice.

src/grid/atomgrid.py
FarnazH: All constructors take `rgrid` as an argument, so like the other two, I thought `from_predefined` should also have it as the first argument. I think it's important to be consistent. I would be happy to change it back if you feel strongly about it.

tczorro: One of the reason I prefer `atomic number` first, because logically, I think it's what pop up naturally in my head if I'm going to generate a grid for certain atom. I'd think I'm gonna generate a grid for some atom, then I would wonder what radial grid should I use to best fit the atom.

The inconsistency comes from the fact that only the `predefined` is `atomic number` sensitive. I got how you feel about the first args changing-inconsistency, but I'd think logic matters more. I will think about how to make it feels better

tczorro: Or we can make it a key-arg at the end of the function call. I am fine with that.

FarnazH: Which argument are you suggesting to turn into a keyword argument? I am a bit confused.

Let's see what others have to say. I mostly wanted to make sure we are making a conscious decision, and this choice isn't an artifact of code legacy. This is a minor issue after all. 

I understand your point. But, from the ease-of-use perspective, I think the user should not need to remember which constructor takes `rgrid` as the 1st argument and which constructor takes `rgrid` as the 2nd argument. Even though 

The `from_pruned` also takes `atomic radius` as an argument, so based on your logic, should that also come first?

PaulWAyers:
Pruned grids should have a specified number of radial points for each atom; the specific choice of radial grid and transformation should be user-specified, though one could argue for using the same default as is usually done. 

So 
1. We need to specify not only sectors and the Lebedev degrees/sector, but the total number of radial points for each atom (in each pruned grid). This tends to be relatively simple rules. The Ochsenfeld pruned grids are a nice example, as they come in a format that is convenient for us.
2. *Optionally* we could also keep track of what the pruned grid from the literature used as an underlying radial grid+transformation. I would tend *not* to do this, but instead to default to the best radial grid (whatever @rayhe88 finds) and use that instead.

src/grid/atomgrid.py
tczorro: Let's make `atnum` as a key-arg then it should be good.

For `from_pruned`, I don't remember the radius of any atoms except H. so it's fairly strange to me to think it in the first place.

PaulWAyers: I'd tend to prefer the same order of arguments, so I guess starting with rgrid. I see Derrick's point, but insofar as rgrid is essential always (and a bit unexpected, at least for me) it's good to start with it. Also, I can imagine a way to use a default atnum if we needed to...I'm not recommending that, but it would be more sensible than a universal choice of radial grid.

tczorro:
@FarnazH Could you update the black style so I can merge the commit asap

PR #117, opened 2021-04-28
Update onedgrid

FarnazH:
This PR fixes docstrings of functions in `onedgrid` to make them complete & consistent. It also adds comments to clarify the computations. This PR is **not** finished yet, but I encountered a tentative bug, so I am opening this PR.

### DIFF
--- a/src/grid/onedgrid.py
+++ b/src/grid/onedgrid.py
@@ -372,51 +372,40 @@ def RectangleRuleSine(npoints):
 
 
 def TanhSinh(npoints, delta=0.1):
-    r"""Generate 1D grid on [-1,1] interval based on Tanh-Sinh rule.
+    r"""Generate 1-D grid on [-1, 1] interval based on Tanh-Sinh rule.
 
     The fundamental definition is:
 
     .. math::
-        \int_{-1}^{1} f(x) dx \approx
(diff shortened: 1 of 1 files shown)


src/grid/onedgrid.py
```diff
@@ -165,221 +177,165 @@ def GaussChebyshevType2(npoints):
...
     if npoints <= 1:
-        raise ValueError("npoints must be greater that one, given {npoints}")
-    idx = np.arange(npoints) + 1
-    points = (2 * idx - 1) / (2 * npoints)
+        raise ValueError(f"Argument npoints must be an integer > 1, given {npoints}")
+
+    points = (2 * np.arange(1, npoints + 1, 1) - 1) / (2 * npoints)
 
     weights = (
         (2 / (npoints * np.pi ** 2))
```
FarnazH, line 341: @rayhe88 & @tczorro: Checking Eq. (6) in **1Dgrids.pdf** (which is also included in the docstring), I think the denominator should be `npoints**2 * np.pi`, is this a bug, or the equations were wrong? If the latter, can you share the reference?

rayhe88: @FarnazH  I'm afraid it is a bug, I checked the equations and the reference again. And yes it is `np.pi * npoints ** 2 `

FarnazH: Opened #120 to fix this bug.

Issue #67, opened 2019-07-08
preloading can be made a bit more efficient


tovrstra:
At the moment, pre-loading of lebedev grids is done inside the `AtomicGrid` class, meaning that it gets repeated for every atom, if I understand correctly (should be tested). It can also be implemented in the lebedev.py module, such that preloaded grids stay in memory until the process ends. (The memory overhead is very small.)

tczorro:
Add global cache variable to store all loaded `lebedev` grid for fast I/O

PR #123, opened 2021-05-20
Add global cache_dict to store loaded npz file #67

tczorro:
1. Add cache to lebedev grid incase repetitive I/O
2. Add cache contents test
3. Remove redundant cache in atomic level

### DIFF
--- a/src/grid/lebedev.py
+++ b/src/grid/lebedev.py
@@ -70,7 +70,7 @@ LEBEDEV_NPOINTS = {
 # Lebedev dictionary for converting grid's degrees (keys) to number of grid points (values)
 LEBEDEV_DEGREES = dict([(v, k) for k, v in LEBEDEV_NPOINTS.items()])
 
-cache_dict = {}
+leb_cache = {}
 
 
 def generate_lebedev_grid(*, degree=None, size=None, cache=True):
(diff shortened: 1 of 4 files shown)


tovrstra (commented):
Looks good. Just a few minor questions:

- I guess you can remove the lines that were commented out, or should they stay around?
- Would you be ok with moving `cache_dict` to `lebedev.py`? I'd be inclined to do so for reasons of modularity, i.e. avoiding top-level imports and keeping things together that belong together. It would also fit well with the dictionaries already present in `lebedev.py`.

tczorro:
Those commented are supposed to be deleted. I will try put the cache_dict inside lebedev

src/grid/lebedev.py
```diff
@@ -70,8 +70,10 @@
 # Lebedev dictionary for converting grid's degrees (keys) to number of grid points (values)
 LEBEDEV_DEGREES = dict([(v, k) for k, v in LEBEDEV_NPOINTS.items()])
 
+cache_dict = {}
 
-def generate_lebedev_grid(*, degree=None, size=None):
+
+def generate_lebedev_grid(*, degree=None, size=None, cache=True):
```
FarnazH, line 76: Can you please update the docstring to include `cache` as well? Thanks.

FarnazH (approved):

PR #129, opened 2021-06-22
Reform onedgrid functions into classes

tczorro:


FarnazH (approved):
Some of the docstrings should be updated/completed, but that has nothing to do with this PR, so we should address it later.

PR #133, opened 2021-07-16
Cubic/Uniform grid class

Ali-Tehrani:
This pull-request is regarding Grids in three dimensions.   There are three classes: _CubicGrid, Tensor1DGrids, and UniformCubicGrid.   Here, I define cubic to be that it is specified by 3 indices (i, j, k) (which I think is usually defined as a cube),  and uniform means consequent points are evenly spaced.   This implies that a parallelepiped is cubic, even though  it is slanted cube.

Tensor1DGrids and UniformCubicGrids are children of the base class _CubicGrid.  This base class contains the interpolation function, which utilizes Cubic interpolate with scipy.   

UniformCubicGrids class is largely copied from [chemtools.utils.grid](https://github.com/theochem/chemtools/blob/master/chemtools/utils/cube.py).  UniformCubicGrids contains four integration weighting scheme from the file [cubicintegrations](https://drive.google.com/file/d/18apJ_2vzZH71kE1pW3CDcIRhG53JyMn5/view?usp=sharing). These weighting schemes are tested with a simple Gaussian (TestUniformCubicGrid.test_integration_with_gaussian) inside test_cubic, and unfortunately, I was only able to achieve two decimal place precision.  The calculation of the integration weighting scheme were vectorized, and the tests compare them to the brute-force method.

Inside chemtools.utils.cube (UniformCubicGrid._calculate_volume), the volume is calculated assuming if the rows in the "axes" attribute are orthogonal to one another, else the volume should be computed like a [parallelepiped](https://mathinsight.org/image/volume_parallelepiped).  I'll need someone else to confirm this.  This assumption of orthogonal axes isn't stated in the docs in chemtools either.


tczorro:
Could you take some time fix the `style` errors.

FarnazH:
@Ali-Tehrani I also noticed that there are no tests. You can get some tests from ChemTools (which tested against HORTON & other packages).

src/grid/cubic.py
```diff
@@ -0,0 +1,516 @@
...
+
+
+class CubicGrid(Grid):
+    r"""
+
+    Attributes
+    ----------
+
+    Methods
+    -------
```
FarnazH, line 14: We have not been listing Attributes/Methods in the docstrings, so don't worry about it.

src/grid/cubic.py
```diff
@@ -0,0 +1,516 @@
...
+
+
+    """
+    def __init__(self, oned_grids):
```
FarnazH, line 18: Isn't it easier and more clear to describe, if 3 `OneDGrid` are passed instead of a list of them? I.e., what about
```
def __init__(self, oned_x, oned_y, oned_z):
```

src/grid/cubic.py
```diff
@@ -0,0 +1,516 @@
...
+        Construct the UniformCubicGrid object.
+
+        Parameters
+        ----------
+        smallest_pnt : ndarray(3,)
+            The smallest point in each axis in the 3D cubic grid.
+        largest_pnt : ndarray(3,)
+            The largest point in each axis in the 3D cubic grid.
+        step_sizes : ndarray(3,)
+            The step-size between two consecutive points in each axis in the 3D cubic grid.
```
FarnazH, line 228: Usually, the user specifies the origin (or the `smallest_pnt`), the `step_size`, and the number of points in each direction. I don't think requiring the `largest_pnt` is that user-friendly.

src/grid/cubic.py
```diff
@@ -0,0 +1,516 @@
...
+            Input integrand array is given or not of proper shape.
+
+        """
+        return super().integrate(*value_arrays)
+
+    def interpolate_function(
+            self, real_pt, func_values, use_log=False, nu_x=0, nu_y=0, nu_z=0,
+    ):
+        r"""
+        Interpolate function at a point.
```
FarnazH, line 93: @tczorro is refactoring the `interpolation` module, so I am wondering whether this is sth that should be added there.   Please check with him @Ali-Tehrani.

src/grid/cubic.py
```diff
@@ -0,0 +1,516 @@
...
+        return self._points[0]
+
+    def __len__(self):
+        """Return the number of grid points."""
+        return self._points.nsize
+
+    def evaluate_function_on_grid(self, f):
+        r"""
+        Evaluates the function f on each point of the cubic grid.
+
```
FarnazH, line 310: Can you please explain why this method is needed @Ali-Tehrani? I am just thinking that we don't have any methods of this kind for other grids, and the points are always passed to a function for evaluation on the grid.

FarnazH (changes requested):

src/grid/cubic.py
Ali-Tehrani: Resolved in 064daac

src/grid/cubic.py
Ali-Tehrani: Resolved in d958179

src/grid/cubic.py
Ali-Tehrani: This is old, I removed it after our meeting last week, resolved in d958179.

src/grid/cubic.py
Ali-Tehrani: Resolved in d958179

src/grid/cubic.py
```diff
@@ -0,0 +1,575 @@
...
+            )
+        self._shape = shape
+        super().__init__(points, weights)
+
+    @property
+    def shape(self):
+        r"""Return number of points in each direction."""
+        return self._shape
+
+    def interpolate_function(
```
FarnazH, line 60: Thanks @Ali-Tehrani for including the interpolation functionality as well. I wonder whether this should be turned into a private method or we should add an interpolation function to other grids as well? I am also wondering whether this should just return the splines? This is because we have an interpolation module, but if I am not wrong, it assumes the given grid is an atom-centered spherical grid and returns spline (to be used for example for an ODE solver)... What are your thoughts @Ali-Tehrani?

What do you think @tczorro? Can you please share a code snippet to show how the refactored interpolation/evaluation module works?

src/grid/tests/test_cubic.py
```diff
@@ -0,0 +1,392 @@
+r"""Cubic Grid Testing."""
+
```
FarnazH, line 2: Can you add the Grid Header to this test file?

FarnazH (commented):
These are my comments for the first round of code reviews. Thanks @Ali-Tehrani

src/grid/tests/test_cubic.py
Ali-Tehrani: Fixed in 4f0e86c

src/grid/cubic.py
```diff
@@ -0,0 +1,575 @@
...
+    def __len__(self):
+        """Return the number of grid points."""
+        return self._points.size
+
+    def _calculate_volume(self, shape):
+        r"""Return the volume of the Uniform Cubic Grid."""
+        volume = np.linalg.norm(shape[0] * self.axes[0])
+        volume *= np.linalg.norm(shape[1] * self.axes[1])
+        volume *= np.linalg.norm(shape[2] * self.axes[2])
+        return volume
```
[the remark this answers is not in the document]
Ali-Tehrani: Fixed in 7b4efcd and simple test is added in ce43302

src/grid/cubic.py
```diff
@@ -0,0 +1,575 @@
...
+        return self._axes
+
+    @property
+    def origin(self):
+        """Cartesian coordinate of the origin of the cubic grid."""
+        return self._origin
+
+    def __len__(self):
+        """Return the number of grid points."""
+        return self._points.size
```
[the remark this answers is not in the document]
Ali-Tehrani: Thanks, I just removed it, it was left-over from QTAIM code. This is fixed in 7b4efcd

src/grid/cubic.py
```diff
@@ -0,0 +1,575 @@
...
+
+
+class UniformCubicGrid(_CubicGrid):
+    r"""Uniform Cubic Grid, a grid whose points are evenly spaced apart in each axes."""
+
+    def __init__(self, origin, axes, shape, weight_type="Trapezoid"):
+        """
+        Construct the UniformCubicGrid object.
+
+        Grid whose points in each (x, y, z) direction has a constant step-size/evenly spaced.
```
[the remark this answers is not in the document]
Ali-Tehrani: Thanks, I added equations for the uniform grid and the integration weight in 9249daa .  When I start the sphinx autodic I'll fix the layout of the equations then.

src/grid/cubic.py
```diff
@@ -0,0 +1,575 @@
...
+        r"""
+        Construct the CubicGrid class, i.e. each point is specified by (i, j, k) indices.
+
+        Parameters
+        ----------
+        points : np.ndarray(M, 3)
+            The 3D points of the cubic grid.
+        weights : np.ndarray(M*3)
+            The weights corresponding to each cubic grid point.  It moves in the following
+            order (x, y, z).
```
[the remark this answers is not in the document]
Ali-Tehrani: Thanks, fixed in d13a530.

src/grid/cubic.py
Ali-Tehrani: Don't think it's worthwhile to add interpolation function to 1DGrids, because of how easy it is (1-2 lines of code). Not sure of other grids.
There wouldn't be universal spline that would work anywhere because interpolation of the x-coordinate would depend on the y-coordinate you want to interpolate etc.  Since it's nested interpolation, the only cubic splines that is used everywhere is a list of cubic spline for each point in the z-axis.  After that, it will use each of the cubic spline in the z-axis to construct another cubic spline in the y-axis that will depend on the point you want to interpolate. But not entirely certain, need to double check.
Don't know much about having a private method because I don't know the use-case here.

src/grid/cubic.py
```diff
@@ -0,0 +1,582 @@
...
+        points = coords.T.dot(self._axes) + origin
+
+        # assign the weights
+        weights = self._choose_weight_scheme(weight_type, shape)
+        super().__init__(points, weights, shape)
+
+    @classmethod
+    def from_molecule(cls, atnums, atcoords, spacing=0.2, extension=5.0, rotate=True):
+        r"""Construct a uniform grid given the molecular coordinates."""
+        raise NotImplementedError
```
FarnazH, line 413: @Ali-Tehrani can you please complete this method? See https://github.com/theochem/chemtools/blob/master/chemtools/utils/cube.py#L98

src/grid/cubic.py
```diff
@@ -0,0 +1,582 @@
...
+
+
+class Tensor1DGrids(_CubicGrid):
+    r"""Tensor1DGrids : Tensor product of three one-dimensional grids."""
+
+    def __init__(self, oned_x, oned_y, oned_z):
+        r"""Construct Tensor1DGrids by tensor product of three one-dimensional grids.
+
+        Parameters
+        ----------
```
FarnazH, line 261: @Ali-Tehrani Can you please complete the docstring? Also, I was thinking that `oned_z` can be optional (with a default value of `None`) in which case a 2D grid is made. Can you please add that?

src/grid/cubic.py
```diff
@@ -0,0 +1,582 @@
...
+        if not isinstance(oned_z, OneDGrid):
+            raise TypeError(
+                f"Argument oned_z should be an instance of `OneDGrid`, got {type(oned_z)}"
+            )
+
+        # number of points in x, y, and z direction of the cubic grid
+        shape = (oned_x.size, oned_y.size, oned_z.size)
+
+        # Construct 3D set of points
+        points = (
```
FarnazH, line 280: @Ali-Tehrani Can you please add a comment here to clarify in which order the points are stored? Is it (z, y, x)?

FarnazH (commented):
I added a few commits to your branch and here are my comments:

FarnazH (commented):
Thanks @Ali-Tehrani for the updates. I fixed a failed test and polished the docstrings some while going through the code. 

I just have one last question: As the `_HyperRectangleGrid` and `Tensor1DGrids` both work in 2D & 3D dimensions, can you have `UniformCubicGrid` work in 2D as well? (I can see this being useful for ChemTools) Feel free to let me know what you think.

Ali-Tehrani:
I have finished 2D uniform grid, had to change the naming from UniformCubicGrid to UniformGrid.   The only thing I didn't implement in 2D from UniformGrid class is the class method `.from_molecule`. 

@FarnazH, @PaulWAyers The last problem I have are is that the "Fourier2" integration weights (Eqn 1.7 from [Cube_integrals.pdf](https://drive.google.com/drive/folders/1ZBL50iFLchFspMYMcaNon_5seCrWAYua))  is incorrect or at-least it doesn't give me the same accuracy I would get from the other integration schemes from that file.  I'm integrating a single Gaussian where I know it is zero at the ends of the cube.  It could be that I implemented the equation wrong but I created a brute-force test and it passed.   Considering that the book is 600 pages, I decided not to look into it further.

PaulWAyers:
> I have finished 2D uniform grid, had to change the naming from UniformCubicGrid to UniformGrid. The only thing I didn't implement in 2D from UniformGrid class is the class method `.from_molecule`.
> 
> @FarnazH, @PaulWAyers The last problem I have are is that the "Fourier2" integration weights (Eqn 1.7 from [Cube_integrals.pdf](https://drive.google.com/drive/folders/1ZBL50iFLchFspMYMcaNon_5seCrWAYua)) is incorrect or at-least it doesn't give me the same accuracy I would get from all other integration schemes from that pdf file. I'm integrating a single Gaussian where I know it is zero at the ends of the cube. It could be that I implemented the equation wrong but I created a brute-force test and it passed. Considering that the book is 600 pages, I decided not to look into it further.

This is from section F.6 in [Boyd's book](https://depts.washington.edu/ph506/Boyd.pdf). I (tried to) do the transformation from (0,pi) to (0,1). 

Equation 1.5 in my notes corresponds to appendix F.4 (equation F.23) in Boyd.

Section F.6 *does not* force the function to decay at the boundary, and so we'd expect it to be a bit less accurate. Comparing Eq. F.32 in Boyd's book to my formula I don't see a mistake, but I could have made a silly error (twice). It is a strange case because the assumption is that the function is antisymmetric, which imposes a sort of zero-value boundary condition, but it isn't taken explicitly here.

Issue #100, opened 2020-07-29
BeckeWeights.generate_weights should raise an exception when it tries to use Nan radii


tovrstra:
The following example will cause the becke partitioning to use Nan radii. The resulting grid gives large integration errors but not `Nans`:

```python
import numpy as np
from grid.becke import BeckeWeights
from grid.molgrid import MolGrid
from grid.onedgrid import GaussChebyshev
from grid.rtransform import BeckeTF
becke = BeckeWeights(order=3)
oned = GaussChebyshev(150)
rgrid = BeckeTF(1e-4, 1.5).transform_1d_grid(oned)
grid = MolGrid.horton_molgrid(np.array([[0, 0, -1], [0, 0, 1]]), np.array([2, 2]), rgrid, 74, becke)
print(grid.weights.min())
print(grid.weights.max())
```

Output:

```
/home/toon/miniconda3/envs/horton3/lib/python3.7/site-packages/grid/becke.py:82: RuntimeWarning: invalid value encountered in greater
  alpha[alpha > cutoff] = cutoff
/home/toon/miniconda3/envs/horton3/lib/python3.7/site-packages/grid/becke.py:83: RuntimeWarning: invalid value encountered in less
  alpha[alpha < -cutoff] = -cutoff
-121784734604746.08
109395391879381.38
```

The resulting grid is not usable, so it would be safer to raise an exception when trying to construct becke weights with atoms for which the radii ar not defined.

tczorro:
add an error to Radon

PR #136, opened 2021-07-19
Fully fix nan radii problem for noble atom #100

tczorro:


src/grid/becke.py
```diff
@@ -162,7 +162,8 @@ def generate_weights(self, points, atcoords, atnums, *, select=None, pt_ind=None
             [
                 self._radii[num]
                 if not np.isnan(self._radii[num])
-                else self._radii[num - 1]
+                else np.nan_to_num(self._radii[num - 1])
+                or np.nan_to_num(self._radii[num - 2])
```
FarnazH, line 166: A comment here would be helpful to clarify how the `np.nan` radii are treated. I like the `else` statement; it's pretty clever. It took me a quick try to see what is happening, so not bad to explain that as well.

PR #137, opened 2021-08-08
Add kwargs to preset for a more user friendly API

tczorro:
Changs in `AtomGrid`
1. Set `rgrid` optional
2. Set `atnum` and `preset` to kwargs only


FarnazH (approved):
Thanks, @tczorro. This looks good to me. I find it a bit strange for the optional `rgrid` argument to appear before the required (keyword-only) arguments like `atnum` & `preset`, but I am not sure there is anything we can do. Also, I think this is not uncommon.

PR #138, opened 2021-08-16
Add spherical harmonics and interpolation

tczorro:
Add function to generate spherical harmonics
Add function to fit radial value into cubic splines
Add function to interpolate points on splines

### DIFF
--- a/src/grid/atomgrid.py
+++ b/src/grid/atomgrid.py
@@ -307,10 +307,15 @@ class AtomGrid(Grid):
         list[scipy.PPoly]
             A list of cubic spline fitted by given value arrays
         """
+        if value_array.size != self.size:
+            raise ValueError(
+                "The size of values does not match with the size of grid\n"
+                f"The size of value array: {value_array.size}\n"
+                f"The size of grid: {self.size}"
+            )
         if self._basis is None:
             theta, phi = self.convert_cart_to_sph().T[1:]
             self._basis = self._generate_real_sph_harm(self.l_max // 2, theta, phi)
-        print(self._basis.shape)
         prod_value = self._basis * value_array * self.weights
         rad_values = [
             np.sum(prod_value[:, self.indices[i] : self.indices[i + 1]], axis=-1)
@@ -327,15 +332,15 @@ class AtomGrid(Grid):
             CubicSpline(x=self.rgrid.points, y=sph_val) for sph_val in ml_sph_values
         ]
 
-    def interpolate(self, points, values, deriv=0):
-        """Interpolate values at given points on the splines fitted by provided values.
+    def interpolate(self, points, value_array, deriv=0):
+        """Interpolate values at given points on the splines fitted by provided value_array.
 
         Parameters
         ----------
         points : np.ndarray(N, 3)
             N points to be interpolated on the splines fitted by given values
             These points are required to be in Cartesian coordinates
-        values : np.ndarray(M,)
+        value_array : np.ndarray(M,)
             a 1d-array evaluabled at each atomic grid point
         deriv : int, optional
             order of derivative to be evaluated on the spline, by default 0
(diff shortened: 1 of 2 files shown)


src/grid/atomgrid.py
```diff
@@ -288,6 +291,108 @@ def convert_cart_to_sph(self, points=None, center=None):
...
+        ]
+        # rad_values in shape (n_shell, n_sph_harms)
+
+        ml_sph_values = np.array(rad_values).T  # shape(n_sph_harms, n_shell)
+        ml_sph_values /= self.rgrid.points ** 2 * self.rgrid.weights
+        return [
+            CubicSpline(x=self.rgrid.points, y=sph_val) for sph_val in ml_sph_values
+        ]
+
+    def interpolate(self, points, splines, deriv=0):
```
FarnazH, line 324: Following `scipy.interpolate` API and the cubic grid's `interpolate` method in PR #133, this should take the `points` and `values` (not `splines`). This would be more user-friendly as well. Can you please fix it?

src/grid/atomgrid.py
```diff
@@ -288,6 +291,108 @@ def convert_cart_to_sph(self, points=None, center=None):
...
+        print(self._basis.shape)
+        prod_value = self._basis * value_array * self.weights
+        rad_values = [
+            np.sum(prod_value[:, self.indices[i] : self.indices[i + 1]], axis=-1)
+            for i in range(self.n_shells)
+        ]
+        # rad_values in shape (n_shell, n_sph_harms)
+
+        ml_sph_values = np.array(rad_values).T  # shape(n_sph_harms, n_shell)
+        ml_sph_values /= self.rgrid.points ** 2 * self.rgrid.weights
```
FarnazH, line 319: Can you add some comments to clarify what you divided by `self.rgrid.weights`?

FarnazH (changes requested):
Thanks, @tczorro for this update.

src/grid/atomgrid.py
tczorro: So to combine the fitting and interpolate functions together?

FarnazH: Well, `interpolate` uses the fitting function under the hood, but we should keep the fitting function. Does that make sense?

tczorro: It makes sense and `fitting` function need to be kept for other functionality. But in this way, everytime the `interpolate` being called, it need to recompute the splines. If that's not a very frequently happened case, I'm fine with the change.

FarnazH: That shouldn't be a problem, as I cannot see the case that it would be called repeatedly, also the `interpolate` method takes an array of points. Passing `points` & `values` makes it easier to use and more consistent (with Scipy and CUbic Grid which cannot return a spline all the time). So, I think it's best to go with it for now.

tczorro: OK. Will change the code accordingly

PaulWAyers: I think it's OK. It's a lot more user friendly to fit to values. Since we support an array of points, I doubt it will often be that we repeatedly need to repeat things. If we preserve low-level functionality with Derrick's previous strategy, then an (expert) user could use that method if (s)he needed to avoid repeated evaluation of splines, while at the (general) user level the intuitive points/values syntax was supported.  Does this make sense @FarnazH @tczorro

tczorro: That makes sense. I have updated the API and put some extra notes/comments. Let me know if any extra modification is needed

src/grid/atomgrid.py
```diff
@@ -288,6 +291,113 @@ def convert_cart_to_sph(self, points=None, center=None):
...
+        ----------
+        value_array : np.ndarray(N,)
+            a 1d-array evaluated at each atomic grid point
+
+        Returns
+        -------
+        list[scipy.PPoly]
+            A list of cubic spline fitted by given value arrays
+        """
+        if self._basis is None:
```
FarnazH, line 307: Can you add the check at the beginning of this method to make sure `value_array` has `(self.size,)` shape?

FarnazH (approved):
This looks good to me; probably the docstrings/comments need more polish but we should do that when everything is in place. Please just add the check below, and I am good to merge. Thanks, @tczorro.

Issue #125, opened 2021-05-27
Bug when trying to use  `MultiExpTF` transformation


leila-pujal:
I tried to use `MultiExpTF` transformation combined with GaussChebyshevType2 radial grid by using the next code: 

```
from grid.onedgrid import GaussChebyshevType2
from grid.rtransform import MultiExpTF

oned = GaussChebyshevType2(30) 
rgrid = MultiExpTF(1e-6, 1.5).transform_1d_grid(oned) 
```

And got this error:
ValueError                                Traceback (most recent call last)
<ipython-input-1-a38563b7d61d> in <module>
      3 
      4 oned = GaussChebyshevType2(30)
----> 5 rgrid = MultiExpTF(1e-6, 1.5).transform_1d_grid(oned)

~/grid/grid/src/grid/rtransform.py in transform_1d_grid(self, oned_grid)
     99         if new_domain is not None:
    100             new_domain = self.transform(np.array(oned_grid.domain))
--> 101         return OneDGrid(new_points, new_weights, new_domain)
    102 
    103     def _convert_inf(self, array, replace_inf=1e16):

~/grid/grid/src/grid/basegrid.py in __init__(self, points, weights, domain)
    256             if len(domain) != 2 or domain[0] > domain[1]:
    257                 raise ValueError(
--> 258                     f"domain should be an ascending tuple of length 2. domain={domain}"
    259                 )
    260             min_p = np.min(points)

ValueError: domain should be an ascending tuple of length 2. domain=[1.e+16 1.e-06]

PR #141, opened 2022-04-12
Refactor documentation in onedgrid and rtransform

Ali-Tehrani:
This pull-request refactors (mostly documentation fixes) for onedgrid.py and rtransform.py.  

Key changes are:
- Added type hinting
- Changed the documentation in both files by fixing typos, adding equations, some references.
- Made it more explicit as to which intervals it's being mapped to in the documentation.
- Fixed flake8 and black issues
- cfcffc214e528b22ef7331ad9b541572c9c9a875 For Trefethen transformation in onedgrid.py, I made it more explicit that the d parameter is the degree of the Taylor polynomial of sin^{-1} and changed it so that it matches the equations, i.e. d=1 is identity, d=3 is the third degree etc. See the paper in [dropbox](https://drive.google.com/drive/folders/1ZBL50iFLchFspMYMcaNon_5seCrWAYua).
- da8509e685df865a5a0e6a0383dc1a39c7b63f8e Changed the naming of TF to RTransform to make it consistent across all classes in rtransform.py
- I changed HortonLinear to UniformInteger, because this generates an equally-spaced integer grid from 0 to N-1 with weights equal to one.  This plays a huge role below.
- The problem in issue #125, is that the MultiExponential transform transform points to `f(x) = -R log ((x + 1) / 2)` and thus it swaps the endpoints `f(1) = 0` and `f(-1)=infty`.  Hence, when the base class OneDGrid transforms, it sees that the bounds [infity, 0] is invalid because it isn't ordered correctly.  I fixed this in 4b17f0a25bd5171c5baeaa27b0041a0d578f8f41 by always sorting the domain to get [0, infty].

The biggest issue I have is the radial transformations (LinearRTransform, ExpRTransform, PowerRTransform, HyperbolicRTransform) which were obtained from [Horton/grid.py]() and how the documentation in python grid isn't correct as to what the domain/codomain are and they seem to be very specific to HortonGrid/UniformInteger class.   For example, for LinearRTransform it transforms according to `r(x) = rmin + x*(rmax - rmin)/(npoint - 1)`, where npoint is the number of points, rmin is the minimum of grid and rmax is maximum of grid.    

The docs claim that it maps to `[r_min, r_{max}]` but this is only true if you're using HortonLinear/UniformInteger grid.  This is also a problem in ExpRTransform and PowerRTransform, where it only maps to `[r_{min}, r_{max}]` only if you're using HortonLinear grid.  One easy way of fixing this, is to replace `npoint - 1` to taking the maximum over all grid points i.e. `np.max(grid.points)`, this will hold true for any grid.

For HyperbolicRTransform, the mapping `r(x) = ax/ (1 - bx)` where a, b are positive scalars. It doesn't map to [0, \infty) but rather it can map everywhere.  Additionally, there is the constraint `b(npoints - 1) >= 1` that isn't explicit in the docs, I'm assuming it is also specific to HortonGrid where it centers the singularity at `npoints - 1` and you only want the positive points on the left-side of singularity, see [desmos/grapher](https://www.desmos.com/calculator/ik0eevj29b), where `N=npoints`.

References
----------------
List of radial transforms:  https://www.pamoc.it/tpc_num_int.html#VOL-B

### DIFF
--- a/src/grid/molgrid.py
+++ b/src/grid/molgrid.py
@@ -165,7 +165,7 @@ class MolGrid(Grid):
 
         Example
         -------
-        >>> onedg = HortonLinear(100) # number of points, oned grid before TF.
+        >>> onedg = UniformInteger(100) # number of points, oned grid before TF.
         >>> rgrid = ExpRTransform(1e-5, 2e1).generate_radial(onedg) # radial grid
         >>> becke = BeckeWeights(order=3)
         >>> molgrid = MolGrid.from_size(atnums, atcoords, rgrid, 110, becke)
--- a/src/grid/ode.py
+++ b/src/grid/ode.py
@@ -20,7 +20,7 @@
 """Generic ode solver module."""
 # from numbers import Number
 
-# from grid.rtransform import InverseTF
+# from grid.rtransform import InverseRTransform
 from numbers import Number
 
 import numpy as np
--- a/src/grid/onedgrid.py
+++ b/src/grid/onedgrid.py
@@ -28,21 +28,26 @@ from scipy.special import roots_chebyu, roots_genlaguerre
 
 
 class GaussLaguerre(OneDGrid):
-    """Gauss Laguerre integral quadrature class."""
+    r"""
+    Gauss Laguerre integral quadrature class.
 
-    def __init__(self, npoints: int, alpha: float = 0):
-        r"""Generate 1-D grid on [0, inf) interval based on Generalized Gauss-Laguerre quadrature.
+    The definition of generalized Gauss-Laguerre quadrature is:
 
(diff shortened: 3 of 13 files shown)


Ali-Tehrani:
The new changes I made are moving the equations from the constructor in the OneDGrid classes in onedgrid.py to the class document ([f4759e1](https://github.com/theochem/grid/pull/141/commits/f4759e1f82689aefdd688f03fb67dccee281952d)) and I copied the transformation and inverse equations and added it to the class docs in rtransform.py ([aba6c7e](https://github.com/theochem/grid/pull/141/commits/aba6c7e008ab0326967f10df7b5254c0617ab41f)).

FarnazH (approved):

PR #146, opened 2022-05-20
Spherical average

Ali-Tehrani:
This pull request adds the spherical average of a function in AtomGrid class.  It returns a splines that takes in radial points.

The tests that are included are:
- The spherical average of a Gaussian is itself and its integral (with 4 pi r^2) of the spherical average function is the integral of the Gaussian (pi^1.5)
- The spherical average of a Spherical Harmonic (whose degree is less than the degree of the atom grid) is zero.


FarnazH (commented):
@Ali-Tehrani, I fixed 2 bugs in the commits above. Feel free to let me know what you think.

FarnazH:
@Ali-Tehrani, I implemented the correct spherical integration instead of taking the mean for small r which are not that many (option 3 in your message). We can talk about whether we want to add a utility to let the user obtain the AngularGrids when we meet.

PR #150, opened 2022-06-10
Add github action for CI

tczorro:
Add github action CI configure
Update tox.ini
Remove poisson.py and tests temporarily as it breaks CI
Update style and docs
Set lowest python test version to 3.7


FarnazH:
@tczorro we have to use `roberto` https://github.com/theochem/roberto for CI so that it makes release on `pip` and `conda` automatic. At the first glance, it didn't look like you are using Rob, am I right?

PR #152, opened 2022-07-25
Refactor poisson.py and ode.py

Ali-Tehrani:
This pull-requests makes changes to poisson.py and adds solving ode as initial value problem for ode.py

ode module
----------------
- Added solving ordinary differential equation as a initial value problem, called solve_ode_ivp (64eac5920f1114de31b42b8257545a10099132b8). I renamed the previous ode method to solve_ode_bvp, since it solves it as a boundary value problem.  The function that takes the solution of these ode and reconstructs the function is added a seperate function (4d614f5f41a32411df2c8e03da3cb63d904af042) since it is used by both solve_ode_ivp and solve_ode_bvp.
- Two tests were added to solve_ode_ivp  (7fedcdd921f6f465f3be103dd76307795babab18):
    - Compared it against analytic examples (without using radial transformation). These analytic examples were solved using woflramalpha.
    - Compared the solution using transformation and without transformation and made sure the results matched each other. The transformation were chosen so that the minimal point `r_{min}` matches the initial point of `x_span`.

Poisson module
---------------------
- Removed the previous Poisson module and added solving poisson as bvp or ivp problem (765ac44a076ce0850f0e621a136ee070de3bc8b9).
     - For `solve_poisson_bvp` function, it has the option of including the `r=0` point and removing large points (useful when radial transformation gives infinite values or 1e6 values).  This greatly helped in increasing accuracy.   The boundary value problem is optional, and if not provided it computes the boundaries instead.
     - For `solve_poisson_ivp` function, the initial value problem is calculated based on the following asymptotic to the solution of the Poisson:
image
The derivative of this gives the initial value for the derivative for large x.
- The following tests were added for Poisson (1d86225e63957706f1f094343614b85a1ccb4cd2)
     - Tested with analytic result for `solve_poisson_bvp` with unit-charge distribution (f(x) is Gaussian) whose solution is error function `erf(r)/r` that is undefined at `r=0`. 
     - Tested the solution from `solve_poisson_bvp` with interpolate_Laplacian, unfotunately the Laplacian  isn't accurate with `r\approx 0` and so it was only tested on region `r\in[0.1, \infty)`.
     - Tested `solve_poisson_ivp` with unit-charge distribution. I have tried many kinds of one-dimensional grids and radial transformation and unfortunately only found one pair that gave accurate results, including the origin!   I've added a comment in the module docs of poisson.py to recommend using bvp if values near the singularity (which is the origin for unit-charge distribution) is needed
- The `interpolate_laplacian` was moved from atomgrid to poisson.py and is accurate away from the origin due to many divisions by `1/r, 1/r^2` and the accuracy of interpolation of the radial-components. I added a cut-off option to remove points close to origin (cdd268512be9469fef466cbf22902cdb734645b2).
     
     
Extra Changes
--------------------
- The one-dimensional grid GaussLaguerre will give nans for weights when the npoints is above a certain point. This is due to scipy implementation. Added a raise error if nans are included (ce5635c078ff5c44b32573227aa5790fb0258348)
- **Nevermind, this wasn't a bug, I made a mistake and reverted it with improved docs in ([82c66c8](https://github.com/theochem/grid/pull/152/commits/82c66c843aff33dce3c4ab467db6c41e68abf6bf))** Fix a bug in finding the radial components inside atomic grid (06d45772a1da6ba75377a54078636d55fed825c2).  General function  expanded in the real spherical harmonic basis:
image where image but this integral is calculated as a Lebedev grid that can only integrate spherical harmonics up to `l_max`. Since multiplication of two spherical harmonics adds their degree, it can only integrate up to `l_max //2`, anything above that is inacurrate and thus set to zero.
- For some radial transformation that can give infinite values, the `trim_inf` attribute wasn't correctly applied and it wasn't added to the first derivatives. This is fixed in (5b5fd47ab487b789ab8bdee38dc6a65264a3be35)


src/grid/poisson.py
```diff
@@ -0,0 +1,419 @@
...
+    is transformed to an set of ODE problems as a initial value problem.
+
+    Ihe initial value problem is chosen so that the boundary of :math:`g` for large r is set to
+    :math:`\int \int \int f(r, \theta, \phi) / r`.  Depending on :math:`f`, this function has
+    difficulty in capturing the origin :math:`r=0` region, and is recommended to keep the
+    final interval :math:`a` close to zero.
+
+    Parameters
+    ----------
+    atomgrid : AtomGrid
```
FarnazH, line 72: Please update so we can support `MolGrid`.

src/grid/poisson.py
```diff
@@ -0,0 +1,419 @@
...
+        r_pts, theta, phi = convert_cart_to_sph(points).T
+        r_values = np.array([spline(r_pts) for spline in splines])
+        r_sph_harm = generate_real_spherical_harmonics(atomgrid.l_max // 2, theta, phi)
+        return np.einsum("ij, ij -> j", r_values, r_sph_harm)
+
+    return interpolate
+
+
+def solve_poisson_bvp(
+    atomgrid: AtomGrid,
```
FarnazH, line 187: Please update so we can support `MolGrid`.

Ali-Tehrani:
The following features were added

- Both Poisson solvers can now handle molecular grids.
- The BVP is a lot more robust than IVP and needs further improvement.

Issue #175, opened 2023-08-03
Python 3.6 installation problems


Ali-Tehrani:
Setuptools version requirement in pyproject.toml is not compatible with Python 3.6:
```
requires = ["setuptools>=65.0", "setuptools_scm[toml]>=7.1.0"]
build-backend = "setuptools.build_meta"
```

Fixing this causes further problems when running `pip install .`  and refers grid to as "UNKOWN":
```
Processing /home/pally/PythonProjects/grid
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Installing backend dependencies ... done
    Preparing wheel metadata ... done
Building wheels for collected packages: UNKNOWN
  Building wheel for UNKNOWN (PEP 517) ... done
  Created wheel for UNKNOWN: filename=UNKNOWN-0.0.5a6.post434-py3-none-any.whl size=3902 sha256=06580d1fd849e27839fde35b780de5cfb1b2c5ef1543a97f6aa1e4feccf50550
  Stored in directory: /tmp/pip-ephem-wheel-cache-d4brkxca/wheels/d1/98/24/14bed4afbf15d42cf2596818cc329de4a11f32aaac3c00d1f3
Successfully built UNKNOWN
Installing collected packages: UNKNOWN
  Attempting uninstall: UNKNOWN
    Found existing installation: UNKNOWN 0.0.5a6.post434
    Uninstalling UNKNOWN-0.0.5a6.post434:
      Successfully uninstalled UNKNOWN-0.0.5a6.post434
Successfully installed UNKNOWN-0.0.5a6.post434
```

The recommended solution is to update setuptools but this didn't work for me. It may be an issue with ubuntu 20.04 and python 3.6.  The easiest solution is to require python>=3.7

PaulWAyers:
There was another package where we recently required Python 3.7 and higher. I'd go with that solution.

PR #180, opened 2023-08-17
Update pyproject.toml to remove Python 3.6 support

PaulWAyers:
Python 3.6 caused issues, see #175 . Deleted this.


Ali-Tehrani:
My bad, didn't realize this was a pull-request

PR #189, opened 2023-09-13
Add documentation on Gaussian quadrature

marco-2023:
Added a Jupyter notebook explaining Gaussian Quadrature.

Missing:
We can make links from here to the different notebooks of the quadrature rules implemented in GRID


FarnazH:
@marco-2023, I am wondering whether the content of this notebook should be added at the beginning of the https://grid.qcdevs.org/pyapi/grid.onedgrid.html module (as a string). This will be like the algorithms described at the beginning of https://grid.qcdevs.org/pyapi/grid.periodicgrid.html module. Feel free to let me know what you think.

marco-2023:
I don't think this is needed. This set of notes doesn't belong in the package.

PaulWAyers:
If you think it is valuable, it could go in the HORTON3 repository, where we have some things showing key workflows and more pedagogical material. Or we could put it in a repository on QC-Edu.

marco-2023:
I will read it with a fresh pair of eyes. Maybe if I clean it, it can be good as pedagogical material. (but this was before I had access to your notes, so they overlap a lot).

PR #191, opened 2023-09-18
General class for N body grid integrations

marco-2023:
Added module and class for generic N grid integrations. This class is the base framework and on top of it, some porcelain will be added for the coulomb-exchange integrals.

### DIFF
--- a/src/grid/__init__.py
+++ b/src/grid/__init__.py
@@ -34,3 +34,4 @@ from grid.ode import *
 from grid.onedgrid import *
 from grid.periodicgrid import *
 from grid.rtransform import *
+from grid.ngrid import *
--- a/src/grid/ngrid.py
+++ b/src/grid/ngrid.py
@@ -1,0 +1,20 @@
+# GRID is a numerical integration module for quantum chemistry.
+#
+# Copyright (C) 2011-2019 The GRID Development Team
+#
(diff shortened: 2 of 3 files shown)


PaulWAyers (approved):
It looks good to me. I'd let @Ali-Tehrani or @FarnazH review it before merging, however.

Can you add a tutorial notebook showing how to integrate the k-electron distribution function from Hartree-Fock (a Slater determinant) with the QC-Devs packages? That notebook could be added in the Horton3 repository, which is where most notebooks that use multiple packages go.

PaulWAyers:
At first I'd just compute normalization of the distribution functions. Then we can work on coulomb integrals in a second step.

marco-2023:
Perfect, I will do so then. Thank you very much.

marco-2023:
@sanchezw17 The Basegrid should not be changed to accommodate the Ngrid class. I will work on an example and see if I can reproduce your problem.

marco-2023:
@sanchezw17  I made the corrections to the ngrid and added some tests. Now It is possible to use grids of different dimensions for the same function. At some point an example notebook will be needed but for now, you can see the tests. Please let me know if you have any problem after the corrections.

marco-2023:
I went through this PR again, still LGTM. I will be merging it.

PR #199, opened 2023-11-02
Atomgrid examples

marco-2023:
Added a notebook showcasing atomic grid capabilities.


FarnazH (changes requested):
@marco-2023 thanks again for putting together such a great example notebook. I enjoyed going through it. I pushed my revisions to your branch. Most of the changes are as described in https://github.com/theochem/grid/pull/202#pullrequestreview-1738111003. Additionally,

1. Can you get the value of electron repulsion from the log file, and include it in the notebook for comparison?
2. Can you add an example for `AtomGrid.radial_component_splines` and `Atomgrid.get_shell_grid` methods. This can be done in a follow-up PR after merging this work.
3. There were cases that the example description in the markdown and the following code-block were not matching (for example, the number of radial grid point and their range were not matching what was used in the code block). I can understand how that happens, especially when you are developing notebooks from scratch, but to make the notebooks sustainable, it is best not to include much details in the markdown and provide that in the comments.
4. In "Example: Calculate Radial Electron Density Distribution", the formula using sum over grid needs to be updated. Also, I like how you have included computations using radial density (see snippet below). Probably, we can also show how to use `AtomGrid.spherical_average` for this example (and get the same result). Your thoughts?
```python
radial_rho = atgrid.integrate_angular_coordinates(dens)
radial_rho *= rgrid.points**2
```

marco-2023:
@FarnazH Thank you very much for the feedback and your suggestions.

Here I modified the example "Example: Calculate Radial Electron Density Distribution" so that `AtomGrid.spherical_average` is also used. 

Regarding 2. I will add the examples in another PR after this is merged.

FarnazH (approved):

PR #202, opened 2023-11-17
Add Quickstart notebook

marco-2023:


FarnazH (approved):
@marco-2023, thanks for your PR. The content and logic of your notebook were great; I working through it. I added my revisions in b734bbc692ec30e1de349cf6e88465121d134d33. Some of the changes were made automatically by my editor, and the changes I made addressed (the list is not complete):

1. The links should be to the webpage not the `rst` files in the repo. Unless I am missing sth, your links didn't work for me.
2. When making an instance of a class like `MolGrid`, it would be good to showcase its attributes by printing some information about the grid. I changed the print statements to do that.
3. It is better for `MolGrid` example to show an example of a molecule (instead of H atom, which is useful but can be confusing as a quick start example). Then, we don't need to remake the MolGrid for H2+ when using the grid for computations.
4. Sometimes the markdown description did not match the code block, so I updated that. For example, Formaldehyde was mentioned in the comments when H2+ was used in the code/markdown.
5. Let's keep our variable names consistent across notebooks and packages. For example, we have been using `rgrid`, `atgrid`, and `oned` (probably we should make a glossary). It is important to stay consistent for easier understanding and use.
6. Sometimes the text in the markdown and comments repeated the same thing. I would keep the comments minimal, so that code is simpler and easier to maintain. I updated the notebook to fix these.
7. I added a small section to check the accuracy of interpolated density against the analytical value.
8. Moved Cubic and Periodic grids to the end of the notebook to move using the grid for computation (integration and interpolation closer to MolGrid section).

After @Ali-Tehrani has reviewed the notebook, we can merge it, and you can make a new PR when the cubic/periodic parts are done.

marco-2023:
@FarnazH Thanks for the review and the comments. The guidelines you shared will help me a lot in the upcoming notebooks!

 I have a doubt regarding the first point. I made the links pointing to the corresponding `rst` files following what was previously done in the `Molecular grid` notebook. The ideas is that the links are generated by sphinx when building the documentation. When I did it on my computer worked. Should we change and use the static links instead?

PR #209, opened 2024-01-03
Modify `get_atomic_grid`to return also indices and add example

marco-2023:
The method `get_atomic_grid` of `MolGrid` now returns a tuple with the atomic/local grid and the indices of the points corresponding to the atoms. An example was added to the Molecular_Grid  example notebook (integrating an atom charge within a molecule).

Notes:
The weights of the atomic grid returned by `get_atomic_grid` correspond to the integration of the whole space (No aim weights are used). Because of this, if the molecular property on the atom-grid points is integrated, the integral approximately corresponds to the integrated property of the whole molecule. If only the atomic property is wanted, the property values need to be multiplied by the aim weights (that can be obtained from MolGrid), see the example.


src/grid/molgrid.py
```diff
@@ -594,17 +594,21 @@ def get_atomic_grid(self, index: int):
...
+        at_indices = np.arange(self._indices[index], self._indices[index + 1])
+        # get atomic grid if stored and the indices within the molecular grid
         if self._atgrids is not None:
-            return self._atgrids[index]
+            return self._atgrids[index], at_indices
         # make a local grid
         pts = self.points[self._indices[index] : self._indices[index + 1]]
         wts = self._atweights[self._indices[index] : self._indices[index + 1]]
-        return LocalGrid(pts, wts, self._atcoords[index])
+        return LocalGrid(pts, wts, self._atcoords[index]), at_indices
```
FarnazH, line 611: @marco-2023, the `LocalGrid` has an optional argument called `indices` (see https://github.com/theochem/grid/blob/master/src/grid/basegrid.py#L329), so this should return:
`return LocalGrid(pts, wts, self._atcoords[index], indices=at_indices)`

The `AtomGrid` also has `indices` attribute (see https://github.com/theochem/grid/blob/master/src/grid/atomgrid.py#L316), but it is different from `at_indices` returned. Any thought on how to make the two returns consistent?

doc/notebooks/Molecular_Grid.ipynb
```diff
@@ -191,37 +198,81 @@
...
+      ],
+      "source": [
+        "# Extract the atomic grid of the Oxygen atom\n",
+        "o_grid, o_idcs = mol_grid.get_atomic_grid(1)\n",
+        "\n",
+        "print(f\"Molecular grid has {len(mol_grid.points)} points.\")\n",
+        "print(f\"Oxygen grid has {len(o_grid.points)} points.\")\n",
+        "\n",
+        "# Extract electron density on the Oxygen grid points\n",
+        "o_electron_density = electron_density[o_idcs] * mol_grid.aim_weights[o_idcs]\n",
```
FarnazH, line 238: @marco-2023, I was thinking that it woul be more straightforward if `mol_grid.aim_weights[o_idcs]` were already included in the `o_grid`. That is, `o_grid` is not just a simple atomic grid, but it is aware of `aim_weights`. Your thoughts? We can talk about this.

FarnazH (commented):
Thanks @marco-2023, we should discuss the `Local properties` example tomorrow.

src/grid/molgrid.py
marco-2023: @FarnazH Thank you very much. I will fix the LocalGrid case, regarding the rest, we can discuss it tomorrow. I am not yet clear on the original intended use of the `get_atomic_grid` method. Regarding using it for integrating atomic /local properties inside a molecule, the are several options. We can decide tomorrow how to do it.

FarnazH:
See PR #213.

Issue #179, opened 2023-08-17
Molecular Grid from Atomic Grids


PaulWAyers:
There should be an easy-to-use functionality to get molecular grids from atomic grids, as well as an overall wrapper to take atomic positions and types and construct atomic grids directly, then molecular grids.

PaulWAyers:
A starting point is to add a notebook that shows how to construct a molecular grid from atomic grids.

Ali-Tehrani:
I agree with adding in the notebook on how to construct it using the initializer but I feel like the __init__ method achieves the easy-to-use functionality if the user has an example

https://github.com/theochem/grid/blob/cdf1ec3d4282421968b3db215417f6fd09f486bc/src/grid/molgrid.py#L92-L98 

There is also an example in the code docs of [molecular grid](https://grid.qcdevs.org/pyapi/grid.molgrid.html)

The latter wrapper will depend on the radial grid and consequently the results of the grid assessment.

FarnazH:
- [ ] Address comments 1 and 2 from https://github.com/theochem/grid/pull/199#pullrequestreview-1738493652
- [ ] Double check the 2nd formula in the “Calculate Radial Electron Density Distribution” section
- [ ] Probably we can identify the maximum radial electron density distribution using scipy

PR #210, opened 2024-01-04
Default Grid Assesment

Ali-Tehrani:
This pull-requests addresses Issue #179.
The Default rgrid was obtained from [Horton](https://github.com/theochem/horton/blob/master/data/grids/tv-13.7-5.txt). This corresponds to using the UniformInteger grid and PowerRTransform with a "fine" preset for the angular component. This is stored in grid/utils.py in a dictionary with the following:
```python
r"""
Obtained from theochem/horton/data/grids/tv-13.5.txt with the folllowing preamble

These grids were created by Toon Verstraelen in July 2013 in a second effort
to generate tuned grids based on a series of diatomic benchmarks computed
with the QZVP basis of Ahlrichs and coworkers. They are constructed to give
on average 5 digits of precision for a variety of molecular integrals, using
the Becke scheme with k=3.
"""
# Items are (rmin, rmax, npts) in angstrom, Meant for PowerRTransform
```

If the user provided rgrid as None, then it generates the default radial grid.
The following functions and their tests (with two added) were changed to have default rgrid options:
- AtomGrid.from_preset
- MolGrid.from_preset
- MolGrid.from_size
- MolGrid.from_pruned


Ali-Tehrani:
Added default BeckeWeights to aim_weights to all static, constructor methods

src/grid/molgrid.py
```diff
@@ -628,3 +654,33 @@ def __getitem__(self, index: int):
...
+
+    Returns
+    -------
+    OneDGrid:
+        One-dimensional grid that was transformed using PowerRTransform.
+
+    """
+    if atnum in _DEFAULT_POWER_RTRANSFORM_PARAMS:
+        rmin, rmax, npt = _DEFAULT_POWER_RTRANSFORM_PARAMS[int(atnum)]
+        rmin, rmax = rmin * 1.8897259885789, rmax * 1.8897259885789
```
FarnazH, line 679: 1. Please use `scipy.constants.angstrom / scipy.constants.value("atomic unit of length")` for unit conversion.
2. Please add a comment that r values are converted from Angstom to Bohr.

FarnazH (changes requested):
Thanks @Ali-Tehrani.
1. Shouldn't we make the `rgrid` in `from_pruned` method optional as well?
2. Thinking more after our discussion, I agree that we should remove `*_` in arguments. It seems to have been used arbitrarily and as we are not using it anywhere else, it is best to remove it for now. Can you please remove them?

src/grid/molgrid.py
Ali-Tehrani: This is fixed in the following comments

Ali-Tehrani:
The *_ is removed.

If you're referring to AtomGrid.from_prune, default radial grid isn't added because atomic numbers is not included in its signature.

FarnazH (approved):
I removed `*` from the rest of AtomGrid class, which caused an API change. If you have no further comments, feel free to merge this PR.

marco-2023:
I did not find a way to add a commit to this PR. The only change that I would  make is to substitute the values in line 198
```
                rmin, rmax = rmin * 1.8897259885789, rmax * 1.8897259885789
```
 by the corresponding scipy constants 

```
                # convert angstrom to bohr
                ang2bohr = scipy.constants.angstrom / scipy.constants.value("atomic unit of length")
                rmin, rmax = rmin * ang2bohr, rmax * ang2bohr
```

just for consistency with the molgrid module.

FarnazH:
> I did not find a way to add a commit to this PR. The only change that I would make is to substitute the values in line 198 of AtomGrid:
> 
> ```
>                 rmin, rmax = rmin * 1.8897259885789, rmax * 1.8897259885789
> ```
> 
> by the corresponding scipy constants
> 
> ```
>                 # convert angstrom to bohr
>                 ang2bohr = scipy.constants.angstrom / scipy.constants.value("atomic unit of length")
>                 rmin, rmax = rmin * ang2bohr, rmax * ang2bohr
> ```
> 
> just for consistency with the molgrid module.

@marco-2023, Ali already did this change in [c6ff8f4](https://github.com/theochem/grid/pull/210/commits/c6ff8f43eac0ef323d4a601ec86532b78ea8e630).

PR #211, opened 2024-01-04
Add cube methods

marco-2023:
Added the methods for generating a cube from a  uniform grid and vice versa.


marco-2023:
TODO: The data values on the points of the cube file are not returned when reading a grid from the cube file as it is now. I need to add this before merging the PR.

FarnazH:
> TODO: The data values on the points of the cube file are not returned when reading a grid from the cube file as it is now. I need to add this before merging the PR.

@marco-2023, I am taking care of this while reviewing the PR.

FarnazH (approved):

PR #221, opened 2024-01-12
Fixed error in real spherical harmonics docstring equation

pujaltes:
I believe there is a small typo in the docstring for the real spherical harmonics in the `generate_real_spherical_harmonics` function. This PR simply adds the missing square root.


Ali-Tehrani:
Thank you for finding and fixing this!

Ali-Tehrani (approved):

Issue #208, opened 2024-01-03
Fix `DeprecationWarning` and `PytestRemovedIn8Warning`


FarnazH:
Running the tests generates a lot of warning messages, which should be addressed.

Ali-Tehrani:
This has been fixed in #229 and been verified [here](https://github.com/theochem/grid/actions/runs/7728511357/job/21069457581)

PR #229, opened 2024-01-30
Fix raising warnings and pytest depreciation (#208)

Ali-Tehrani:
This pull-requests fixes:
- that pytest removed the option to capture warnings via None, this gives an error when using Python 3.11, see [here](https://github.com/scikit-learn/scikit-learn/issues/22572) and the our pull-request that failed recently #228.  This is heavily used in `test_periodic`, 
- Further `Nose` is used in periodic and is removed by me due to reasons specified by [pytest](https://docs.pytest.org/en/stable/deprecations.html#support-for-tests-written-for-nose).
- Handles all warnings given in all of the tests. This handles issue #208. 
    - For the UserWarnings, I used `pytest.warn(match=)` to make sure the warnings were raised. 
    - In cases, where the warnings were coming from other modules I ignored them use the decorator `@pytest.mark.filterwarnings("ignore:api v1")` to ignore it on the specific test or if I couldn't find out which test were raising a warning I used `pytestmark = pytest.mark.filterwarnings("ignore:...")` to ignore the entire module. This are all specified in the [pytest docs](https://docs.pytest.org/en/7.1.x/how-to/capture-warnings.html).
    - Ignore division by zero and nan in our ode or Poisson solver.
- Added vectorization of interpolation derivatives in cubic with a simple fix. This removes the cubic grid warning.
- Fixed the overflow in spherical harmonics by moving the square-root and adding numpy longdouble precision (just in case) and added a test for orthonormality in high dimensions.

### DIFF
--- a/src/grid/cubic.py
+++ b/src/grid/cubic.py
@@ -102,7 +102,7 @@ class _HyperRectangleGrid(Grid):
         return x, y
 
     def interpolate(self, points, values, use_log=False, nu_x=0, nu_y=0, nu_z=0, method="cubic"):
-        r"""Interpolate function value at a given point(s).
+        r"""Interpolate function and its derivatives on cubic grid.
 
         Only implemented in three-dimensions.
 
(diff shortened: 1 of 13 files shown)


FarnazH (approved):
Thanks @Ali-Tehrani for all your fixes and improvements.

PR #230, opened 2024-01-30
Fix colab badges

marco-2023:
@Ali-Tehrani HTML-style hyperlinks are not supported by nbsphinx so the google-colab badges don't work when rendered in the readthedocs pages. This PR provides a Fix for that. The solution is taken from https://github.com/QData/TextAttack/issues/244#issuecomment-674476435

One question:
Should we make an "Installation" notebook (separated from the paper notebook) so that we have a "complete" tutorial?

### DIFF
--- a/examples/Angular_grid.ipynb
+++ b/examples/Angular_grid.ipynb
@@ -4,9 +4,7 @@
       "cell_type": "markdown",
       "metadata": {},
       "source": [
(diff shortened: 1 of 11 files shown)


Ali-Tehrani:
Thanks Marco, looks good to me

Ali-Tehrani (approved):

PR #239, opened 2024-02-01
Fix Optional Arguments

FarnazH:
Please see the commits and let me know if you have any questions.

PR #240, opened 2024-02-01
Make argument names consistent, update docstring, & add units

FarnazH:
- Update the docstring of the argument center and specify atomic units
- Arguments sectors_degree and deg_sectors were used to refer to degree sectors. so they were both changed to d_sectors.
- sectors_r and radius_sectors were used to refer to radius sectors, so they were both changed to r_sectors. Unit was clarified in docstring.


marco-2023:
The changes are good, but now several tests are failing (I will update them).

marco-2023:
@FarnazH I had to use defaults (`None`) for the `degree` and `size` arguments of the `AngularGrid` ` __init__ `. If no default argument is given for `degree`, then,  when initializing from `size`, it is necessary to provide a `degree=None`. This was breaking many of the tests. Please feel free to change this (it just was the only way that I could find to solve this issue).

FarnazH:
> @FarnazH I had to use defaults (`None`) for the `degree` and `size` arguments of the `AngularGrid` `__init__`. If no default argument is given for `degree`, then, when initializing from `size`, it is necessary to provide a `degree=None`. This was breaking many of the tests. Please feel free to change this (it just was the only way that I could find to solve this issue).

@marco-2023, the `degree` should be set to `None` by the user if they want to make a grid using the `size`. That's the API change I proposed in PR #239, which is merged. In my opinion, it does not make sense to set the default values of `None` for `degree` and `size` when at least one of them should be provided. I understand that we want to support two ways of initializing the grid, but the concept of optional arguments is misused. As it is more common to use `degree`, the user can set it to `None`, if they choose to use `size`.

We can talk more about this today, as this is an important API change appearing in other classes as well.

FarnazH:
@Ali-Tehrani and @marco-2023, the notebook failure can be ignored for now. We are changing the API, and just updating the  `JCP_paper.ipynb` that is needed for submitting the paper. Then we can go back and fix the other tutorial notebooks.

Ali-Tehrani:
- The changes made here don't work for python 3.7, so I added the fixes in commit [afb9a8e](https://github.com/theochem/grid/pull/240/commits/afb9a8e8c293b039ca5335a8f914c8d94108d269).  Also added python 3.7, 3.8 to github actions to make it more evident.  
- Fix the ruff linter complaints with using `Union[a, b]` to `a | b`. 
- Fixed the importlib problem with using Python =3.7.

Ali-Tehrani:
My bad , I didn't realize running JCP_paper was installing the master branch of grid and so I was constantly getting errors since the changes in the Pull-request weren't made. I think this is good to merge and will fix most of the problems.

Ali-Tehrani (approved):

PR #268, opened 2025-02-14
Refactor ngrid

marco-2023:
Refactored the ngrid module to improve readability and add features.

- Added weights property which returns the weights for each argument combination (due to its combinatorial nature returns a generator)
- Added points property which returns the combinations of arguments (due to its combinatorial nature returns a generator). Each point is a tuple with several arrays, each array corresponds to an argument of the multigrid function to integrate.
- Added support for non-vectorized functions. These will be evaluated point by point during the integration.

@PaulWAyers I think this PR improved the previous class. Nonetheless, I still have some doubts:
1. The current class used is a derived class of `Grid` because of this, several methods are inherited `moments`, `save` and `get_localgrid`. I think that `save` and `get_localgrid` can be implemented (I have some doubts about the meaning of `get_localgrid` in this context ), but I don't see how to get `moments`. 
2. This one is related to my previous doubt. Wouldn't be better to make a new class called perhaps `MultiDomainGridIntegrator` that is only responsible for integrating functions with more than one argument?

### DIFF
--- a/src/grid/ngrid.py
+++ b/src/grid/ngrid.py
@@ -104,32 +104,37 @@ class MultiDomainGrid(Grid):
         else:
             return np.prod([grid.size for grid in self.grid_list])
 
-    def integrate(self, callable, **call_kwargs):
-        r"""
-        Integrate callable on the N particle grid.
+    @property
+    def weights(self):
+        """Generator: Combined weights of the multi-dimensional grid.
 
-        Parameters
-        ----------
-        callable : callable
-            Callable to integrate. It must take a list of arguments (one for each particle) with
-            the same dimension as the grid points and return a float (e.g. a function of the form
-            f([x1,y1,z1], [x2,y2,z2]) -> float).
-        call_kwargs : dict
-            Keyword arguments that will be passed to callable.
-
-        Returns
-        -------
-        float
-            Integral of callable.
-        """
-        # check that grid_list is not empty
-        if len(self.grid_list) == 0:
-            raise ValueError("The list must contain at least one grid")
-
-        if len(self.grid_list) == 1 and self.num_domains is not None and self.num_domains > 1:
-            return self._n_integrate(self.grid_list * self.num_domains, callable, **call_kwargs)
+        Due to the combinatorial nature of the grid, the weights are returned as a generator"""
+        if len(self.grid_list) == 1 and self.num_domains is not None:
+            # Single grid repeated for multiple domains
+            weight_combinations = itertools.product(
+                self.grid_list[0].weights, repeat=self.num_domains
+            )
         else:
-            return self._n_integrate(self.grid_list, callable, **call_kwargs)
+            weight_combinations = itertools.product(*[grid.weights for grid in self.grid_list])
+
+        # Yield the product of weights for each combination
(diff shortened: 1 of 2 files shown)


PaulWAyers:
Thanks @marco-2023 . I'm going to add @FarnazH to the review panel for this one.

Ali-Tehrani:
- I would recommend writing out what the items (and their dimension) of the generators are from `points()` and `weights()` function, because I'm kinda confused. I'm guessing its `(x_1, x_2, x_3, \cdots, x_N)`  as a list[list], where $x_i$  is a $n_i$-dimensional vector and N is the number of domains.
    * In addition, I would recommend writing out `f: \mathbb{R}^{n_1} \times \cdots \times \mathbb{R}^{n_N} \rightarrow \mathbb{R}`. I'm pretty certain most if not all, fields understand this notation and makes it clear for me at-least. 
    
- If you're going to do a high-dimensional moments function, then centers should be `centers = List[List[Points]] (C, N, n_i)` where C is the number of  centers.

    * Since there isn't any use-cases of any of the spherical harmonics moments, raise a `NotImplementedError(This has not been implemented, please make an Issue on the Grid page and we'll get back to you.)` or something. Probably good to comment that Spherical Harmonics github page and this pull-request there as well.

   * Radial moments 
 $$\int_{ \mathbb{R}^{n_1} \times \cdot \times \mathbb{R}^{n_N}  } || \mathbf{r}  - R_c ||^{k}  f(\mathbf{r})$$, 
the norm is based on product space of vector-spaces and so the norm is $\sqrt{\sum_{i=1}^N ||x_i||_{\mathbb{R}^{n_i}}^2 }$, where $\mathbf{r} = (x_1, \cdots, x_N)\in \mathbb{R}^{n_1} \times \cdot \times \mathbb{R}^{n_N}$ and I'm guessing  $k = 0, \cdots, M$ with $M = n_1 + \cdots + n_N$.

       - I think the code must be easy, Concatenate all the centers together to get shape (C, N * n_i), then  feed that into moment function in BaseGrid class  `super().moments(orders, centers, func_vals, "radial") `. I could be wrong though because of `self.weights`, but the code should be identical so you can copy-paste and use `MultiDomainGrid.integrate`.

    * Cartesian moments,   similarly to above this, if you set it up correctly, it should work by just feeding it into the Base class super().moments().  But you'll have to change the `moments` so that `orders: int, list` so the user can arbitrary put in the orders as a list.  The code would break when `type_mom = (pure, pure-radial)` but should work as a list for `cartesian` and `radial`.       Or you can copy-paste, the code should work in N-dimensions.
    
        - Generation of the Cartesian orders for high-dimensions might be too much (you'll have to generate Cartesian orders per $\mathbb{R}^{n_i}$), and would probably be better to let the user specify exactly what order they want  
    
I am not guaranteeing  if these suggestions would make it easy or not.  I think you'll have to change `self.weights` as well within BaseGrid.moments, or copying the BaseGrid.moments into MultiDomainGrid and modify it. 
My personal preference is to just  raise a NonImplementedError, since no-one has a good use case and tell the user to make a issue.

PaulWAyers (approved):
LGTM

PR #273, opened 2025-03-05
Update cubic.py

shubhamachary56821:
done some minor changes for  fixing negative shape in Gaussian Cube, if issue still persist, please inform and guide

### DIFF
--- a/src/grid/cubic.py
+++ b/src/grid/cubic.py
@@ -706,10 +706,11 @@ class UniformGrid(_HyperRectangleGrid):
         if not fname.endswith(".cube"):
             raise ValueError("Argument fname should be a cube file with *.cube extension!")
 
-        with open(fname) as f:
+        with open(fname, "r") as f:
             # skip the title and second line
-            f.readline()
-            f.readline()
+            lines = f.readlines()
+            # f.readline()
+            # f.readline()
 
             def read_grid_line(line):
                 """Read a number and (x, y, z) coordinate from the cube file line."""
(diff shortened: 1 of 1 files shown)


src/grid/cubic.py
```diff
@@ -726,8 +727,19 @@ def read_grid_line(line):
             shape0, axis0 = read_grid_line(f.readline())
             shape1, axis1 = read_grid_line(f.readline())
             shape2, axis2 = read_grid_line(f.readline())
-            shape = np.array([shape0, shape1, shape2], int)
-            axes = np.array([axis0, axis1, axis2])
+            # shape = np.array([shape0, shape1, shape2], int)
+            # axes = np.array([axis0, axis1, axis2])
+            # Detect Bohr vs. Angstrom units
+            BOHR_TO_ANGSTROM = 0.529177
```
Ali-Tehrani, line 733: I recommend extending the precision to 16 digits, since our calculations often rely on 64-bit floating point representations. 
Using `scipy` and `mpmath`,  the value for converting 1 Bohr to Angstrom is `0.529177210902999962`.

But since we have SciPy as a prerequisite, i think it's better to do 
```
import scipy.constants as spc
BOHR_TO_ANGSTROM = spc.value('atomic unit of length') / spc.angstrom
```

Ali-Tehrani (changes requested):
Thank you for the pull-request and for implementing this feature!
My only other comment is that
`                "atcoords": coordinates,` will also need its unit changed and the method documentation should include that its units are converted to atomic units (Bohr) length.

Ali-Tehrani:
@PaulWAyers  I think that it is more straightforward to just return a flag, like `units: "bohr" ` (or "angstrom"),  to `cube_data` instead of defaulting to converting everything to atomic units. 

If user specifically cares about atomic units, then they can handle the conversion themselves.  For example I know in protein-related work, they would prefer to keep the units that they are in.

PR #290, opened 2026-02-04
Fix cubic file unit detection for negative dimensions

### LINKED ISSUE #260: Gaussian Cube File Shape
Working with proteins generates Gaussian cube files that are in angstrom. In particular, the pdb2pqr tool ["dx2cube"](https://pdb2pqr.readthedocs.io/en/v3.3.0/using/index.html#dx2cube)  converts the volumetric data from an electrostatic calculation into a Gaussian cube file.

This reports the shape in negative numbers, because it is in Angstrom. See the Gaussian cubegen documentation here: 

Image

This results in a error using `Grid.cubic.UniformGrid.readcube` because it assumes the shape is positive.
The easy fix is to take the absolute value of the shape, but I think it's important to return a flag indicating what units Bohr or Angstrom it is in.

Ao-chuba:
This PR fixes the unit detection logic in `UniformGrid.from_cube`. 
As discussed in the issue, there is a discrepancy between the Gaussian documentation and observed behavior:
- **Documentation**: States `N1 < 0` implies Bohr.
- **Observed (dx2cube)**: Generates files with `N1 < 0` that are in Angstroms.
- **Observed (cubegen)**: Generates files with `N1 > 0` that are in Bohr.

This PR aligns the reader with the observed behavior of tools like `dx2cube` and `cubegen`:
- `N1 < 0` → **Angstrom**
- `N1 >= 0` → **Bohr**

**src/grid/cubic.py**: Updated `UniformGrid.from_cube` to set `unit='angstrom'` when `shape0 < 0`.
**src/grid/tests/test_cubic.py**: Added `test_read_negative_shape` to verify that negative dimensions are correctly interpreted as Angstroms.

closes #260

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -18,7 +18,7 @@ jobs:
       matrix:
         # os: ["ubuntu-latest", "windows-latest", "macos-latest"]
         os: ["ubuntu-latest", "windows-latest"]
-        py: ["3.9", "3.10", "3.11", "3.12", "3.13"]
+        py: ["3.10", "3.11", "3.12", "3.13"]
 
     steps:
       - uses: "actions/checkout@v6"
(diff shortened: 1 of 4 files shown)


marco-2023:
We are doing that on #291

PR #291, opened 2026-02-06
Fix spherical harmonics for new SciPy sph_harm_y API

marco-2023:
Replace deprecated sph_harm with sph_harm_y and update argument ordering to match SciPy’s azimuth/polar convention. Ensure magnetic quantum numbers are always integers, and construct negative-m real harmonics from the imaginary part of positive-m modes.

Simplify HORTON-2 m ordering and fix derivative evaluation to avoid invalid float m values. This restores compatibility with SciPy ≥1.17 and fixes test failures.

### DIFF
--- a/src/grid/utils.py
+++ b/src/grid/utils.py
@@ -573,7 +573,7 @@ def generate_real_spherical_harmonics_scipy(l_max: int, theta: np.ndarray, phi: 
         # generate order m=negative real spherical harmonic: sqrt(2) * (-1)^m * Im(Y_l^m)
         neg_real_sph = (
             np.sqrt(2)
-            * (-1) ** np.abs(m_list_p[:, None])
+            * (-1) ** m_list_p[:, None]
             * sph_harm_y(
                 l_val, m_list_p[:, None], phi, theta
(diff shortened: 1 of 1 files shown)


marco-2023:
> I'm going to approve this, but I still think we should contemplate refactoring to use sph_harm_y_all

At the beginning, I wanted to migrate to the new way of computing the spherical armonics changing as little as possible. But I agree with refactoring towards `sph_harm_y_all` instead. I will take care of it here.

Ao-chuba:
@marco-2023 
Just to confirm - PR #291 will be merged first, right? That should resolve the SciPy related errors currently showing up in #290. 
Do I need to take any action for PR #290, or should I wait for the changes to be updated from your side first?

PaulWAyers (approved):
LGTM

PaulWAyers:
@marco-2023 maybe make a new issue for the refactor to sph_harm_y_all. It can even be a (small) GSoC project, but it is also probably really easy for you or @Ao-chuba to do it at this stage, knowing the code as y'all do.

PR #293, opened 2026-02-10
Add lattice rules for (hyper)cubic grid integration

### LINKED ISSUE #241: Lattice rules for (hyper)cubic grids
Among the most efficient ways to integrate of a (hyper)cube are the [lattice rules](https://link.springer.com/book/10.1007/978-3-031-09951-9). There are several choices here, but probabilty the [embedded lattice rules](https://www.maths.unsw.edu.au/sites/default/files/amr06_35_0.pdf) are the easiest to implement, and they have [been tabulated](https://web.maths.unsw.edu.au/~fkuo/lattice/index.html).

For a (simpler) lead reference, I recommend the [early paper from Sloan](https://epubs.siam.org/doi/epdf/10.1137/0724010).

It would be good to use lattice rules as an alternative for integration over a cube or (after an affine transformation) an arbitrary parallelepiped. I'm much less sure that higher-dimensions are supported, but it is easy to support arbitrary-dimensional integration on a parallelepiped (beyond $d=3000$, which is more than enough!) with a lattice rule.  Interpolation is also relatively straightforward, as it is a generalization of the current structure with specially selected vectors.

Ao-chuba:
This PR implements lattice rules for efficient numerical integration over (hyper)cubic grids, addressing issue #241 .

Following the references provided in the issue ([Sloan's paper](https://www.maths.unsw.edu.au/sites/default/files/amr06_35_0.pdf) and the [UNSW lattice rules database](https://web.maths.unsw.edu.au/~fkuo/lattice/)), I've implemented rank-1 lattice rules as a new integration method.
**Implementation :**

- New Lattice class using tabulated generating vectors from [UNSW database](https://web.maths.unsw.edu.au/~fkuo/lattice/)
- Supports embedded lattice rules (powers of 2, up to 2²⁰ points)
- Equal-weight integration with formula: $\mathbf{x}_k = { \frac{k \cdot \mathbf{z}}{n} }$
- Works with arbitrary parallelepipeds via affine transformation
- Supports up to 50 dimensions (extensible to 3600+ if needed)
- Includes basic interpolation via `scipy.interpolate.LinearNDInterpolator`

**Testing:**

- Input validation, point generation, integration accuracy
- Embedded property verification, interpolation, file I/O
Screenshot 2026-02-11 023415
closes #241

### DIFF
--- a/src/grid/__init__.py
+++ b/src/grid/__init__.py
@@ -29,6 +29,7 @@ from grid.becke import *
 from grid.cubic import *
 from grid.hirshfeld import *
 from grid.angular import *
+from grid.lattice import *
 from grid.molgrid import *
 from grid.ode import *
 from grid.onedgrid import *
--- a/src/grid/lattice.py
+++ b/src/grid/lattice.py
@@ -1,0 +1,307 @@
+# GRID is a numerical integration module for quantum chemistry.
+#
+# Copyright (C) 2011-2019 The GRID Development Team
+#
+# This file is part of GRID.
+#
+# GRID is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 3
+# of the License, or (at your option) any later version.
+#
+# GRID is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, see <http://www.gnu.org/licenses/>
+# --
+r"""Lattice Rules for Integration on (Hyper)Cubic Grids."""
+
+import numpy as np
+from scipy.interpolate import LinearNDInterpolator
+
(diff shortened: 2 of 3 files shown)


PaulWAyers:
@marco-2023 we should maybe talk about this next week?

Ao-chuba:
The Windows CI failure (`Python 3.12 on OS windows-latest`) is **not** caused by this PR.
its failing in `test_angular.py::test_orthogonality_of_spherical_harmonic_at_high_degrees`  the cause being The test sets `degree = 88 * 2`, calling `generate_real_spherical_harmonics(l_max=88)` with 15,756 grid points. At [`utils.py:655`](https://github.com/theochem/grid/blob/master/src/grid/utils.py#L655):

> spherical_harm = np.zeros(((l_max + 1) ** 2, numb_pts), dtype=np.longdouble)`

as for lattice test all pass.
image

@PaulWAyers @marco-2023

Ao-chuba:
After looking into it more  the CI failure was actually caused by the additional tests I added in that commit specifically `test_error_convergence` which created large grids consuming extra memory on the Windows runner. The angular test was right at the memory boundary, and the additional tests tipped it over. After reverting, the CI is green again.

PR #300, opened 2026-04-27
Phase out tox

marco-2023:
Grid was using `pyproject.toml` and `tox.ini` as duplicated sources of truth for package configuration. 
This PR sets  `pyproject.toml` as the only source of truth.


PaulWAyers (approved):
Thanks Marco.

Issue #288, opened 2026-02-02
GSoC 2026: Improve Robustness of Poisson Solver


PaulWAyers:
## Description

Improve the robustness of the Poisson solvers that grid uses to evaluate the Coulomb potential due to a charge distribution,

$$
\Phi(\mathbf{r}) = \int \frac{\rho(\mathbf{r}')}{|\mathbf{r} - \mathbf{r}'|}  d\mathbf{r}'
$$

As pointed out by [Becke](https://pubs.aip.org/aip/jcp/article/89/5/2993/96723/Numerical-solution-of-Poisson-s-equation-in), this is more efficiently treated, numerically, as a solution to the corresponding Poisson equation.

$$
\nabla^2 \Phi(\mathbf{r}) = - 4 \pi \rho(\mathbf{r})
$$

## :books: Package Description and Impact

`Grid` is a pure Python library for numerical integration, interpolation and differentiation of interest for the quantum chemistry community. 

## :construction_worker: What will you do?

`Grid` has two different Poisson solvers, one using a boundary-value method (which seems more robust, but is not perfect #215) and one using an initial-value method which is less robust (see #162). You will attempt to find robust default parameters for these methods. It is conceivable that one needs (much) larger grids for solving the Poisson equation than we routinely use in numerical integration. (Our Gaussian-quadrature-ish grids are especially adapted to integration; for differentiation their accuracy is diminished. Solving the Poisson equation is essentially a differentiation.)

One approach, which would potentially make the method more robust, would be to use a screened Coulomb kernel instead; that approach can remove the singularity in the equation and, by taking the limit to zero screening, give robust results.

Another approach, which might be much easier, would be to use the strategy proposed in #16. That is, estimate the density with a linear combination of Gaussians using [BFit](https://github.com/theochem/BFit), so that one only needs to solve the Poisson equation for the *error* in the fit.

## :checkered_flag: Expected Outcomes

1. Explore the boundary value solver. Find good values for the default parameters. 
2. See if the initial value problem can be made more robust or, failing that, if circumstances where it is robust can be identified.
3. Develop a protocol where a sufficient grid, algorithm, and algorithm parameters can be selected, allowing high-level porcelain to be designed for the function.
4. Explore whether using the "promolecular" trick is beneficial.
5. Explore whether screened Coulomb potentials behave better. 
6. Write comprehensive tests and documentation for all new functionality.
7. Write tutorial Jupyter notebooks that show how to use the new functionality. 

| | |
[...]

PR #304, opened 2026-05-27
PR1 :  implement analytical Coulomb potential for Gaussian densities

Ao-chuba:
First inital PR for Improvement of the robustness of poisson equation . 
This PR implements the core mathematical engine for evaluating the exact analytical electrostatic (Coulomb) potential of $s$-type and $p$-type radial Gaussian charge densities. The math is derived from the exact formulations and $r \to 0$ limits outlined in the course project reference (https://qchem.qc-edu.org/problems/HF_dft.html). This PR contains the standalone `coulomb.py` module and an isolated unit test suite validating math correctness across various distances and normalization schemes, establishing the foundation for the subsequent robust Poisson solver scaffolding.
related to issue -> https://github.com/theochem/grid/issues/288

### DIFF
--- a/src/grid/coulomb.py
+++ b/src/grid/coulomb.py
@@ -72,11 +72,14 @@ def coulomb_gaussian_s(r: np.ndarray, alpha: float, normalized: bool = True) -> 
     if alpha <= 0:
         raise ValueError(f"Gaussian exponent alpha must be strictly positive; got {alpha}")
     r = np.atleast_1d(np.asarray(r, dtype=float))
-    sqrt_a = np.sqrt(alpha)
+    if np.any(r < 0):
+        raise ValueError("Radial distances r must be non-negative")
+
     out = np.empty_like(r)
-    np.divide(erf(sqrt_a * r), r, out=out, where=r >= _R_ZERO_THRESHOLD)
+    sqrt_alpha = np.sqrt(alpha)
+    np.divide(erf(sqrt_alpha * r), r, out=out, where=r >= _R_ZERO_THRESHOLD)
     # safe division
-    out[r < _R_ZERO_THRESHOLD] = 2.0 * sqrt_a / np.sqrt(np.pi)
+    out[r < _R_ZERO_THRESHOLD] = 2.0 * sqrt_alpha / np.sqrt(np.pi)
 
     if normalized:
         return out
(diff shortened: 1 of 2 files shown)


Ao-chuba:
as per copilot suggestions i took into consideration for some of its inputs:
- Added validation inside coulomb_potential ensuring that `coeffs_p, expons_p, and centers_p `are either all provided or all omitted to prevent silent dropping of p-type density.
-  Added checks to ensure that the coefficients, exponents, and centers have the` same length`. This prevents silent truncation of Gaussians due to `Python's zip behavior`.
-  Added checks to ensure` alpha > 0` and` beta > 0`. This prevents invalid math operations (like taking `np.sqrt of negative numbers or division by zero)`.
- changed number `1e-12` to a named module-level constant` (_R_ZERO_THRESHOLD`) to avoid duplication and improve readability.

src/grid/coulomb.py
```diff
@@ -0,0 +1,235 @@
...
+    mask_nonzero = ~mask_zero
+
+    r_nonzero = r[mask_nonzero]
+    sqrt_a = np.sqrt(alpha)
+
+    if normalized:
+        out[mask_nonzero] = erf(sqrt_a * r_nonzero) / r_nonzero
+        out[mask_zero] = 2.0 * sqrt_a / np.sqrt(np.pi)
+    else:
+        prefactor = (np.pi / alpha) ** 1.5
```
marco-2023: The code can and should be shorter and simpler (read):
```python
out = erf(sqrt_a * r) / r
out[r < _R_ZERO_THRESHOLD] = 2.0 * sqrt_a / np.sqrt(np.pi)
```

Code is repeated inside the if-else. It is better to compute out and if not normalizes multiply by the prefactor at the end (you can even do it with return)
```python
# Apply normalization factor only if needed
if normalized:
    return out

return prefactor * out
```

Ao-chuba:
```python
prefactor = 1.0 if normalized else (np.pi / alpha) ** 1.5
out[mask_nonzero] = prefactor * erf(sqrt_a * r_nonzero) / r_nonzero
out[mask_zero]    = prefactor * 2.0 * sqrt_a / np.sqrt(np.pi)
```
Would be better if its structured like this instead?

marco-2023: I don't think we should have two masks just for one or two points. My idea was something like:
```
out = erf(sqrt_a * r) / r
out[r < _R_ZERO_THRESHOLD] = 2.0 * sqrt_a / np.sqrt(np.pi)

if normalized:
    return out

return prefactor * out
```

marco-2023: This pattern is even better
```
np.divide(erf(sqrt_a * r), r, out = out, where=r >= _R_ZERO_THRESHOLD)
out[r < _R_ZERO_THRESHOLD] = 2.0 * sqrt_a / np.sqrt(np.pi)

if normalized:
    return out

return prefactor * out
```

src/grid/coulomb.py
```diff
@@ -0,0 +1,204 @@
...
+    Returns
+    -------
+    np.ndarray
+        Coulomb potential evaluated at the radial distances.
+    """
+    r = np.asarray(r, dtype=float)
+    out = np.empty_like(r)
+    mask_zero = r < 1e-12
+    mask_nonzero = ~mask_zero
+
```
marco-2023: This function can also be simplified, similar to the previous one. 
Compute term 1 and term 2, and use if-else for prefactors at return.
We can also get rid of masks.

src/grid/tests/test_coulomb.py
```diff
@@ -0,0 +1,85 @@
...
+from numpy.testing import assert_allclose
+from scipy.special import erf
+
+from grid.coulomb import coulomb_gaussian_p, coulomb_gaussian_s, coulomb_potential
+
+
+def test_coulomb_gaussian_s():
+    # Test r=0 limit
+    assert_allclose(
+        coulomb_gaussian_s(0.0, 2.0, normalized=True), 2.0 * np.sqrt(2.0) / np.sqrt(np.pi)
```
marco-2023: It is a better practice to store the reference value in one variable, the result in another, and compare them at the end. And being consistent with this across tests, similar to what you did for `test_coulomb_gaussian_p`

marco-2023 (changes requested):
Nice work! I think this is a good progress.
Please simplify the functions.
Homogenize the tests

I think this module should not be in the root of the package, but we can leave that for another moment.

Ao-chuba:
Thanks for the input sir, i updated my code as per the suggestions. please do check and let me know if any other things can be improved too.

marco-2023 (approved):
@Ao-chuba, I added some small changes. It was faster to do than another PR review. I will be merging this pR.

I think more tests are needed (e.g. test behaviour with invalid arguments).
In the future, I think the functions should be vectorized.

Ao-chuba:
Thank you @marco-2023 for the changes. also about the tests, I will keep a note on that alongside the directory at which this module  should be place for future.

PR #305, opened 2026-05-29
Fix Trefethen(Strip)General

merreeves:
TrefethenGeneral and TrefethenStripGeneral OneDGrids can take any OneD Quadrature and transform it. Some OneDGrids have extra parameters which the Trefethen transforms don't account for (e.g. TanhSinh has an additional parameter delta: float = 0.1). The added code corrects this and have tested it.

### DIFF
--- a/src/grid/onedgrid.py
+++ b/src/grid/onedgrid.py
@@ -897,7 +897,7 @@ class TrefethenGeneral(OneDGrid):
 
     name = "Trefethen-Polynomial"
 
-    def __init__(self, npoints: int, quadrature: OneDGrid, d=9):
+    def __init__(self, npoints: int, quadrature: OneDGrid, d=9, **quadrature_params):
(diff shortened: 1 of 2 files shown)


marco-2023:
@merreeves and @PaulWAyers, in my opinion, the way that `TrefethenGeneral` is initialized (from a `OneDGrid` trype and the number of points )  is obscure and fragile. It would be way easier to just give a grid instance. I have a doubt, is there a specific reason why we don't treat `TrefethenGeneral` as a transformation?

PaulWAyers:
I don't think there is any reason not to treat (both of) the Trefethen tricks as a transformation. As I recall, the *mathematical* argument is pretty tightly tied to Chebychev/Fejer/Clenshaw-Curtis quadrature, but there is no real reason to not generalize.

PaulWAyers (approved):
LGTM

PR #312, opened 2026-06-18
Refactor _HyperRectangleGrid coordinates_to_index method to support an arbitrary number of dimensions

marco-2023:


PaulWAyers (approved):
It looks OK to me.

Ao-chuba:
refactor logic looks good to me, just that looking at the co-pilot suggestion i think we need to address those suggestions. 
so we can maybe :
Either scope the docstring to say `"2D/3D"` to match the actual restriction in `__init__,` or open a follow up to remove the `__init__ ` dimensionality check.
Add an if `len(indices) != self.ndim` check at the top of the method, and cast the result to int to be safe.
cc @PaulWAyers @marco-2023

Ao-chuba (approved):

PaulWAyers:
@marco-2023 I think you can merge this now. I'll let you do the honours....

PR #316, opened 2026-06-29
Fix incorrect parameter names in docstrings

eeshsaxena:
### Description

Three docstrings documented parameter names that do not match their function signatures:

- `NumericalIntegrand.integrate` (`src/grid/ngrid.py`) documented `integrand`, but the parameter is `integrand_function`.
- `_evaluate_coeffs_on_points` (`src/grid/ode.py`) documented `coeffs`, but the parameter is `coeff`.
- `deriv3` (`src/grid/rtransform.py`) documented `array`, but the parameter is `x`.

This corrects the documented names to match each signature. Documentation-only change.

### DIFF
--- a/src/grid/ngrid.py
+++ b/src/grid/ngrid.py
@@ -176,7 +176,7 @@ class MultiDomainGrid(Grid):
 
         Parameters
         ----------
-        integrand : callable
+        integrand_function : callable
             Integrand function to integrate. It must take a list of arguments (one for each domain)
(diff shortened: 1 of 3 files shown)


eeshsaxena:
Good catch on the duplicate `integration_chunk_size` - I removed the stray first entry (it had the wrong default of 1000 and appeared before `non_vectorized`); the remaining entry matches the signature's 6000 default and parameter order.

marco-2023 (approved):

PR #321, opened 2026-07-28
Fix Typos

FarnazH:
Used Codex to fix typos or bad grammar in source code and non-Python files.


FarnazH:
@marco-2023, I will merge this as it does not need review.

FarnazH:
@marco-2023, I will merge this as it does not need review.

PR #323, opened 2026-08-15
Add Halton low-discrepancy sequence grid

Zohaibcode740:
## Summary

* Add a multidimensional Halton low-discrepancy sequence grid.
* Generate dimensions using successive prime bases and radical inverses.
* Assign uniform integration weights.
* Export `Halton` from the `grid` package.
* Add tests covering points, weights, shape, and invalid arguments.
* Add a tutorial notebook demonstrating Halton-based integration over a cubic domain.

## Testing

* Dedicated Halton tests: 4 passed.
* Full test suite: 602 passed, 1 skipped.


Zohaibcode740:
@tovrstra @matt-chan @CestDiego @stevenvdb review it

Samuel-Vangu:
Hi @Zohaibcode740 ,

 below is my detailed review. It's a bit long, but please read it all --
I described everything I thought was important. Let me know if you have
any other questions or suggestions.

Samuel-Vangu:
Hi @Zohaibcode740, here's a detailed review of your PR:

## 1. Bug: `__getitem__` is not implemented

Your class doesn't override `__getitem__`, so it falls back to the one
inherited from `Grid`:

```python
def __getitem__(self, index):
    return self.__class__(self.points[index], self.weights[index])
```

This assumes any subclass can be reconstructed directly from a
`(points, weights)` pair -- but `Halton.__init__` expects
`(npoints: int, ndim: int)` instead, so indexing breaks.

You can reproduce it with:

```python
from grid.halton import Halton

h = Halton(npoints=10, ndim=2)
print("Halton successfully constructed:", h.points.shape)

print("Trying h[3]...")
sub = h[3]  # raises an error
print("This line is never reached")
```

```text
Halton successfully constructed: (10, 2)
Trying h[3]...
Traceback (most recent call last):
  File "...\error.py", line 7, in <module>
    sub = h[3]
  File "...\basegrid.py", line 110, in __getitem__
    return self.__class__(np.array([self.points[index]]), np.array([self.weights[index]]))
  File "...\halton.py", line 58, in __init__
    raise ValueError(f"Argument npoints must be a positive integer, given {npoints}")
ValueError: Argument npoints must be a positive integer, given [[0.125      0.44444444]]
```

**Fix:** override `__getitem__` in `Halton` yourself.

## 2. Compared against `scipy.stats.qmc.Halton`

Good news -- the results match. The only difference is that your sequence
is indexed starting at 1 (i = 1, ..., n), while scipy's implementation (and
the other QMC samplers I added) start at 0 (i = 0, ..., n-1).

## 3. No `randomize`/`seed` parameters

Looks like scrambling wasn't implemented, which is understandable -- it's
genuinely tricky to get right. Rather than implementing it from scratch,
I'd suggest wrapping `scipy.stats.qmc.Halton` instead, the same way I did
for `Sobol`. It already supports randomization via its `scramble` argument,
with two extra benefits: less code to maintain and debug, and better
performance.

## 4. Match the structure used in the Sobol PR

Since scipy's Halton (like its Sobol) already handles randomization, I'd
suggest mirroring the structure I used there:

- Add `origin`/`axes` parameters, to map points onto an arbitrary
  parallelepiped.
- Rename `npoints`/`ndim` to `n_points`/`dimension`, for consistency with
  the other samplers.
- Expose the constructor arguments as read-only properties, e.g.:

```python
@property
def n_points(self):
    """int: Number of points in the Sobol design."""
    return self._n_points

@property
def dimension(self):
    """int: Dimension of the Sobol grid."""
    return self._dimension
```

## 5. License header

One more convention worth following: every `.py` file (`halton.py`,
`test_halton.py`, etc.) should start with:

```python
# GRID is a numerical integration module for quantum chemistry.
#
# Copyright (C) 2011-2019 The GRID Development Team
#
# This file is part of GRID.
#
# GRID is free software; you can redistribute it and/or
# modify it under the terms of the GNU General Public License
# as published by the Free Software Foundation; either version 3
# of the License, or (at your option) any later version.
#
# GRID is distributed in the hope that it will be useful,
# but WITHOUT ANY WARRANTY; without even the implied warranty of
# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
# GNU General Public License for more details.
#
# You should have received a copy of the GNU General Public License
# along with this program; if not, see <http://www.gnu.org/licenses/>
# --
r"""Tests for Sobol' Sequences."""
```

## 6. Notebook errors

The notebook seems to raise some errors when executed -- worth fixing
before merge. Feel free to take a look at the notebooks in my other PRs for
inspiration on structure.

Zohaibcode740 (commented):
I reviewed all

Zohaibcode740 (commented):
Thanks for the detailed review. I went through your comments and pushed updates addressing the issues I identified, including the duplicate import and the requested Halton changes.

I’ve also reviewed the remaining points you mentioned. Please take another look at the latest changes and let me know if anything is still missing or needs further adjustment.

Samuel-Vangu:
Hi @Zohaibcode740, here's a follow-up review of your latest changes:

1. `origin`/`axes` mapping doesn't match the library's convention

`n_points`/`dimension` renamed, properties exposed, `origin`/`axes` added --
good progress. But the mapping itself uses `points @ axes.T`, which doesn't
match the convention used everywhere else in the library (`Lattice`,
`Sobol`, `Kronecker`, `RandomizedLatinHypercube` all use `points @ axes`,
no transpose). Tested directly on a non-symmetric `axes`:

```
With axes.T (your version):        [1.333, 1.]
With axes alone (convention used):  [1.,    1.5]
IDENTICAL? False
```

This isn't a style preference -- for any non-symmetric `axes` (i.e. most
real cases), your version places points at genuinely different
coordinates than every other sampler in the library would, given the same
`axes`. Please drop the `.T` to stay consistent.

And also don't forget to adapt the test file when you're done with those changes.

2. Argument names and order should match the rest of the library

`Sobol`, `Kronecker`, and `LatinHypercube` all use:

```python
def __init__(self, n_points, dimension, seed=None, randomize=True, origin=None, axes=None):
```

Yours currently has `scramble` instead of `randomize`, defaults it to
`False` instead of `True`, and orders `origin`/`axes` before
`seed`/`scramble`. Please rename `scramble` to `randomize`, flip the
default to `True` to match, and reorder the arguments to
`(n_points, dimension, seed=None, randomize=True, origin=None, axes=None)`.

3. Remove the old useless files such as  notebook and PDF report

The old `Halton_grid` notebook and `Halton_Grid_report.pdf` are still in
the PR -- please remove both now that `Halton_Grid_final.ipynb`
supersedes them.

Samuel-Vangu:
@Zohaibcode740 

would you have time to take a look at my PRs (Sobol, Kronecker, LatinHypercube)
when you get a chance? You'd mentioned earlier you'd review them once
available, as the maintainers had asked. 

Just a friendly reminder since none have been looked at yet.

Zohaibcode740 (commented):
All checked and done to merge

Zohaibcode740:
@Samuel-Vangu @tovrstra kindly check it again and review it

PR #329, opened 2026-09-03
Use a deterministic initial guess in solve_ode_bvp

Specter842:
## What

`solve_ode_bvp` seeded its default `initial_guess_y` with `np.random.rand(order, x.size)`, so any call that doesn't pass an explicit guess depends on global `np.random` state.

`solve_poisson_bvp` / `solve_poisson_ivp` never pass a guess, so **every Poisson solve is non-reproducible run to run** — it can converge or hit `max_nodes` depending on RNG state, and the `atol=1e-2` Poisson tests are flaky by construction (no seeding in the test suite).

## Change

- Replace the random default with an all-zeros guess. The ODE solved here is **linear**, so the initial guess cannot change the converged solution — only mesh adaptation — so this is safe.
- Update the docstring.
- Add `test_solve_ode_bvp_default_initial_guess_is_deterministic`: the same linear BVP solved under two different global RNG states now gives bit-identical results.

## Testing

`pytest src/grid/tests/test_ode.py src/grid/tests/test_poisson.py` passes locally. Callers can still pass `initial_guess_y` explicitly to override.

### DIFF
--- a/src/grid/ode.py
+++ b/src/grid/ode.py
@@ -210,7 +210,9 @@ def solve_ode_bvp(
         See `scipy.integrate.solve_bvp` function for more info.
     initial_guess_y : ndarray(K, N), optional
         Initial guess for :math:`y(x), \cdots, \frac{d^{K} y}{d x^{K}}` at the points `x`.
-        If not provided, then random set of points from 0 to 1.
+        If not provided, an all-zeros guess is used. Since this routine only solves
+        linear ODEs, the initial guess does not change the converged solution; it only
+        affects mesh refinement. A deterministic default makes the solver reproducible.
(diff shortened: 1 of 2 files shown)


marco-2023:
@Specter842 thanks for your contribution. As it is, the core of this PR is simply to change the default initial guess from `np.random.rand(...)` to `np.zeros(...)`, this is defensible. The rest adds a lot of explanation and testing (fluff) around. Much of it feels AI-generated. I would strongly prefer a minimal PR that makes that change and if needed a test to cover it. I will reject this PR.

Specter842:
I understand thank you for your time.

Issue #315, opened 2026-06-25
Parameter Study & Benchmarking Results for solve_poisson_bvp and solve_poisson_ivp


Ao-chuba:
@marco-2023 Heres the detailed parameter study and benchmark result that was asked for. 

# Parameter Study & Benchmarking Results

To establish reliable default parameters and compare the performance of the BVP and IVP Poisson solvers, I benchmarked both solvers against a normalized s-type Gaussian charge density:

$$
\rho(r)=\left(\frac{\alpha}{\pi}\right)^{3/2}e^{-\alpha r^2}
$$

which has the analytical electrostatic potential:

$$
\phi(r)=\frac{\mathrm{erf}(\sqrt{\alpha}r)}{r}
$$

The benchmark script used to generate these results is located on my fork branch -->[ here](https://github.com/Ao-chuba/grid/blob/master/examples/parameter_study_poisson.py)

---

# 1. BVP Solver Benchmarks (`solve_poisson_bvp`)

## Setup

- **Radial Grid:** `GaussLegendre(n)`
- **Transform:** `BeckeRTransform(1e-5, R=1.5)`
- **Geometry:** Single center at the origin
- **Angular Degree:** `29`

---

## Study 1 — Effect of Radial Grid Size (`n_radial`)

**Parameters**

- `tol = 1e-6`
- `remove_large_pts = 10.0`
- `α = 0.5`

| n_radial | Relative L2 Error | Max Absolute Error | Time (s) | Status |
|---------:|------------------:|-------------------:|---------:|:------|
| 30 | 2.54e-04 | 1.85e-04 | 4.2 | OK |
| 50 | 3.72e-05 | 2.69e-05 | 4.7 | OK |
| 80 | 6.21e-06 | 4.47e-06 | 5.0 | OK |
| 100 | 2.72e-06 | 1.95e-06 | 5.0 | OK |
| 150 | 5.41e-07 | 3.88e-07 | 5.0 | OK |
| 200 | 1.72e-07 | 1.23e-07 | 5.4 | OK |

### Observations

- Approximately fourth-order convergence is observed.
- `n_radial = 100` provides the best trade off between accuracy and computational cost.
- Increasing the radial grid beyond 100 points yields diminishing returns.

---

## Study 2 — Effect of ODE Solver Tolerance (`tol`)

**Parameters**

- `n = 100`
- `remove_large_pts = 10.0`
- `α = 0.5`

[...]

Issue #215, opened 2024-01-05
ODE didn't converge in bvp poisson solver


ccastilloo:
I had been trying to use `solve_poisson_bvp` to compute electronic repulsion between fragments, but apparently, the ode solver didn't converge and printed me the following error: 
![Captura de pantalla de 2024-01-05 10-44-45](https://github.com/theochem/grid/assets/119383214/a6cbc3c6-5a01-4f95-bbac-d9cb767f44af)

if I remove the origin it works but becomes extremely slow. I'm testing it using a water dimer and I attached the code with their respective files
 
[ode_issue.zip](https://github.com/theochem/grid/files/13842695/ode_issue.zip)

Specter842:
Tried to reproduce this on current master (not the original water dimer, just synthetic 2-center Gaussian densities at various separations/sharpness). With `include_origin=True`, which the solver now defaults to, I couldn't get a non-convergence error at sep=1.4-4.0 and alpha up to 1000; all cases converge with rel-L2 error in the 1e-4-1e-5 range. `include_origin=False` is actually worse now (rel-L2 ~0.3), which is the opposite of what's described here.

Given how much the solver has changed since 2024 (BeckeRTransform defaults, remove_large_pts, etc.), is this still reproducible with the original water-dimer case, or can this be closed / folded into the #263 effort?

PaulWAyers:
@Ao-chuba and @marco-2023 did you confirm that this example works now? 

@Specter842 if you would like to write a tutorial Jupyter notebook for this example, showing how it now works, that would be really a very helpful contribution to the repository. (I don't think it is redundant with other contributions, but @Ao-chuba would know best.)

Ao-chuba:
@PaulWAyers and @Specter842 , I ran the original water dimer files from `ode_issue.zip` on current master using diff approaches.

**Setup:** `MolGrid.from_preset(..., preset="medium")`, `GaussLegendre(100)`, `BeckeRTransform(1e-5, R=2.0)` , slightly different from the original script's `preset="insane"` / `BeckeRTransform(0.0, R=2.0)`, but comparable.

[...]

PR #331, opened 2026-09-05
Enable two-center case in test_poisson_bvp_gives_the_correct_laplacian

Specter842:
This case was commented out with `# TODO: Couldn't get the following test to pass.` I ran it against current master (two Gaussians at (0,0,0) and (1,1,1), default params) and it passes: max abs error ~6e-3 on grid points and ~3e-14 on random points, both under the existing atol=1e-2. Looks like it was fixed as a side effect of other work since this was written, just never re-checked. Uncommenting it.

### DIFF
--- a/src/grid/tests/test_poisson.py
+++ b/src/grid/tests/test_poisson.py
@@ -236,11 +236,7 @@ def test_poisson_bvp_on_unit_charge_distribution(oned, tf, remove_large_pts, cen
         [zero_func, np.array([[0.0, 0.0, 0.0]])],
         [gauss, np.array([[0.0, 0.0, 0.0]])],
         [gauss, np.array([[1.0, 0.0, 0.0]])],
-        # TODO: Couldn't get the following test to pass.
-        # [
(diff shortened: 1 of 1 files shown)


PaulWAyers (commented):
It looks like a good contribution to me. I'd want @marco-2023 and/or @Ao-chuba to confirm that this test now no longer needs to be commented out, however, as I'm not especially familiar with this part of the code.

Specter842:
Gentle bump. CI is green on all platforms here, including the case that was commented out. Happy to add anything that would help confirm it's safe to enable.

Ao-chuba:
@Specter842 
Based on my  benchmarking work #315, the BVP solver has been improved since then. we confirmed this when testing the original water dimer from #215 and the  BVP solver now handles it  without crashing, so enabling this two-center Laplacian test is consistent with all of this. If CI is green, I think it is safe to merge.
also there is no logic change in the code so its a minimal change but correct

Ao-chuba (approved):

PR #332, opened 2026-09-05
Add tutorial notebook: Poisson on a molecular grid

Specter842:
@PaulWAyers suggested this in #215.

A short `examples/Poisson_on_a_Molecular_Grid.ipynb` that runs `solve_poisson_bvp` on a `MolGrid` (two s-type Gaussians on the x-axis, 2 electrons), checks the potential against the analytic sum-of-erf solution (rel L2 ~4e-4), and plots it along the bond axis. Also adds the `_toc.yml` entry so it shows up in the website build.

The existing `Interpolation_and_Poisson.ipynb` only covers the single-atom case, so this fills in the multi-center usage that #215 was about. Executes clean under `jupyter execute` in ~20s.


Specter842:
Tried to redo this against the actual water-dimer wavefunction (matching the `Molecular_Grid.ipynb` pattern: `iodata.load_one` + `gbasis.evaluate_density`) instead of the synthetic two-Gaussian density it currently has.

Couldn't get there cleanly, so leaving this PR as-is for now rather than push something unverified:

- `grid.coulomb.load_atomic_gaussian_params` (the H/O fits already shipped in the repo) looked like a zero-dependency way to build a water-dimer promolecule, but it's the core-subtraction density used by `solve_poisson_robust`, not a full-atom density - my test density integrated to 21.15 electrons instead of 20, so that's the wrong object and I'm not going to fake it.
- The real path needs `iodata`/`gbasis` + an actual water-dimer `.fchk`/wavefunction file, and I don't have one - the repo only ships formaldehyde's. `qc-iodata` also won't install in my current environment (Python 3.14), so I can't verify a gbasis-based notebook locally even if I had the file.

@Ao-chuba since you ran the original `ode_issue.zip` files - do you have the water-dimer wavefunction/geometry from that, or a preferred way to generate one? Happy to build the real version against it.

